PR #3, opened 2019-02-14
Adding 2379 port for openshift for the contracts for service catalog

gaurav-dalvi:


puppet-py:
can you provide a more relevant name, instead of kube-api3. Otherwise LGTM

PR #6, opened 2019-02-19
Prefer infra-vlan configured on APIC

mandeepdhami:
Prefer infra-vlan configured on APIC if it conflicts the infra-vlan
configured in the input YAML file

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -202,19 +202,19 @@ def config_discover(config, prov_apic):
     if prov_apic is not None:
         apic = get_apic(config)
 
+    orig_infra_vlan = config["net_config"].get("infra_vlan")
     ret = {
         "net_config": {
-            "infra_vlan": None,
+            "infra_vlan": orig_infra_vlan,
         }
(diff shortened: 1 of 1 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -202,19 +202,19 @@ def config_discover(config, prov_apic):
     if prov_apic is not None:
         apic = get_apic(config)
 
+    orig_infra_vlan = config["net_config"].get("infra_vlan")
```
gaurav-dalvi: LGTM. I hope value setting to None wont create problem in the code below where it is referenced.

gaurav-dalvi:
Do we need to change files here : https://github.com/noironetworks/acc-provision/tree/master/provision/testdata

provision/acc_provision/acc_provision.py
mandeepdhami: There is a use case where None is a valid value, so we need to handle that. And the following code should handle it.

snaiksat:
We can enhance UTs as a follow up patch.

PR #9, opened 2019-02-21
Functionality to get release info from acc-provision

apoorva11029:
A file inside the acc-provision package called RELEASE-VERSION contains package release info, the git commit ID and the build time.

To get release info, run
acc-provision --release

### DIFF
--- a/provision/setup.py
+++ b/provision/setup.py
@@ -1,9 +1,16 @@
+import sys
+sys.dont_write_bytecode = True
+
 from setuptools import setup, find_packages
+import os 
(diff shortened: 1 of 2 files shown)


provision/setup.py
```diff
@@ -1,9 +1,16 @@
+import sys
+sys.dont_write_bytecode = True
+
 from setuptools import setup, find_packages
+import os 
+file_dir = os.path.dirname(__file__)
```
gaurav-dalvi: This is good. Hopefully, It will pass rpm and deb job this time with this chance.

Issue #23, opened 2019-02-28
Docker-ucp flavor should open additional ports


jojimt:
443, 12378-12388, 12376, 10250
These ports need to be open from kube node to kube-system

gaurav-dalvi:
PR : https://github.com/noironetworks/acc-provision/pull/28

PR #28, opened 2019-03-05
Adding ports to kube-api filter for DockerUCP flavor

gaurav-dalvi:
Resolves : https://github.com/noironetworks/acc-provision/issues/23

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -304,7 +304,7 @@ class ApicKubeConfig(object):
         update(data, self.opflex_cert())
 
         update(data, self.l3out_tn())
-        update(data, getattr(self, self.tenant_generator)())
+        update(data, getattr(self, self.tenant_generator)(self.config['flavor']))
         for l3out_instp in self.config["aci_config"]["l3out"]["external_networks"]:
             update(data, self.l3out_contract(l3out_instp))
 
@@ -1494,7 +1494,7 @@ class ApicKubeConfig(object):
         else:
             return True
 
-    def kube_tn(self):
+    def kube_tn(self, flavor):
         system_id = self.config["aci_config"]["system_id"]
         tn_name = self.config["aci_config"]["cluster_tenant"]
         vmm_name = self.config["aci_config"]["vmm_domain"]["domain"]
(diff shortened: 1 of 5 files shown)


puppet-py:
Suggesting will it be apt to move the list of ports(single_ports) into flavors.yaml both for Openshift & DockerUCP. Since flavors.yaml is one the config entry point (like as we are reading the vmm_type), in future we can add/delete new/existing ports by editing flavors.yaml. This help us to avoid mucking the apic_provision code.

provision/acc_provision/apic_provision.py
```diff
@@ -3456,5 +3412,113 @@ def cloudfoundry_tn(self):
...
+            )
+        ]
+    )
+
+    data['fvTenant']['children'][7]['vzFilter']['children'].append(os_extra_port)
+
+
+def open_port_for_dockerucp(data):
+
+    single_ports = ["443", "12376", "10250", "12378", "12379", "12380", "12381",
```
jojimt: Can we split this into single ports and ranges?

gaurav-dalvi: Done

provision/acc_provision/apic_provision.py
```diff
@@ -3456,5 +3412,116 @@ def cloudfoundry_tn(self):
...
+                                        ),
+                                        (
+                                            "dFromPort",
+                                            port,
+                                        ),
+                                        (
+                                            "dToPort",
+                                            port,
+                                        ),
+                                        (
```
jojimt: @gaurav-dalvi what I meant was the dFromPort and dToPort should be the range. This way, the number of filter entries in ACI will be smaller.

gaurav-dalvi: @jojimt : Got it. Please take a look now.

provision/acc_provision/flavors.yaml
```diff
@@ -173,6 +173,10 @@ flavors:
     default_version: 1.9
     status: Experimental
     hidden: false
+    config:
+      aci_config:
+          vmm_domain:
+            type: DockerUCP
```
snaiksat: On the APIC side we don't have a DockerUCP VMM domain. So I am wondering why this was added? Would it the VMM domain not default to Kubernetes if we don't add this? I believe the intention is to make this available as part of Kubernetes VMM domain.

gaurav-dalvi: i dont think so. DockerUCP is different than K8s. So I have added new VVM_type for it.

snaiksat: We are not creating a new VMM for DockerUCP.

provision/acc_provision/flavors.yaml
```diff
@@ -103,6 +103,8 @@ flavors:
         kubectl: oc
         system_namespace: aci-containers-system
       aci_config:
+        port_range:
+          - [2379, 2379]
```
snaiksat: Its great that you were incorporated this change in response to the review comment. But since we are doing this, we should also enhance this to be able to specify other attributes of the filter entry. Can we do something like this:
aci_config:
    ports:
        - name: openshift-svc-catalog
           range: [2379, 2379]
           etherT: ip
           prot: tcp
           stateful: no
        - name: blah
           range: [p1, p2]
           ...

snaiksat: The formatting was lost in the comment. "range" should be aligned under "name", and similarly for other attributes.

snaiksat:
@gaurav-dalvi  Thanks for incorporating all the review comments.

provision/acc_provision/apic_provision.py
```diff
@@ -3225,54 +3225,13 @@ def kube_tn(self):
...
-                                        ]
-                                    ),
-                                )
-                            ]
-                        ),
-                    )
-                ]
-            )
-            data['fvTenant']['children'][7]['vzFilter']['children'].append(os_extra_port)
+        print(self.config["aci_config"])
```
puppet-py: "print" should be replaced by log methods, be it info or debug

provision/acc_provision/apic_provision.py
```diff
@@ -3325,7 +3284,7 @@ def contract(self, name, subjects=[]):
             children.append(subj)
         return aci_obj("vzBrCP", [('name', name), ('_children', children)])
 
-    def cloudfoundry_tn(self):
+    def cloudfoundry_tn(self, flavor):
```
puppet-py: flavor is not getting handled inside the cloudfoundry_tn. It should be defaulted to None

PR #32, opened 2019-03-07
Override /usr/local in debian packaging 

apoorva11029:


provision/debian/rules
```diff
@@ -21,3 +21,5 @@ override_dh_auto_test:
 	#testr init && ./run_tests.sh -N -P
 	true
 endif
+
```
gaurav-dalvi: LGTM. Please make sure DEB job passes with this change.

PR #34, opened 2019-03-12
Adding openshift-catalog contract, filter and subject for openshift service catalog.

gaurav-dalvi:
Adding openshift-catalog contract, filter and subject for open shift service catalog. Removing the previous addition which was in kube-api contract

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -3225,13 +3225,12 @@ class ApicKubeConfig(object):
                     }
                 })
 
-        print(self.config["aci_config"])
         if "ports" in self.config["aci_config"].keys():
             ports = self.config["aci_config"]["ports"]
             if vmm_type == "OpenShift":
-                open_port_for_flavor(data, ports)
+                openshift_flavor_specific_handling(data, ports)
             elif flavor == "docker-ucp-3.0":
-                open_port_for_flavor(data, ports)
+                dockerucp_flavor_specific_handling(data, ports)
 
         return path, data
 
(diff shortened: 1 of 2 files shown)


provision/acc_provision/apic_provision.py
```diff
@@ -3415,7 +3414,223 @@ def cloudfoundry_tn(self, flavor):
...
+                            ],
+                        ),
+                    ]
+                ),
+            )
+        ]
+    )
+    data['fvTenant']['children'].append(misc1)
+
+    # add filter and entires to that filter
```
snaiksat: should this be:
# add filter and entires to that subject

provision/acc_provision/apic_provision.py
```diff
@@ -3415,7 +3414,223 @@ def cloudfoundry_tn(self, flavor):
         return path, data
 
 
-def open_port_for_flavor(data, ports):
+def openshift_flavor_specific_handling(data, ports):
+    if ports is None or len(ports) == 0:
+        err("Error in getting ports for flavor")
+
+    # add new contract
+    provider_os_contract = collections.OrderedDict(
```
snaiksat: this should probably be:
provide_os_contract

provision/acc_provision/apic_provision.py
```diff
@@ -3415,7 +3414,223 @@ def cloudfoundry_tn(self, flavor):
...
+                                ]
+                            ),
+                        )
+                    ]
+                ),
+            )
+        ]
+    )
+
+    consumer_os_contract = collections.OrderedDict(
```
snaiksat: this should probably be:
consume_os_contract

provision/acc_provision/apic_provision.py
```diff
@@ -3415,7 +3414,223 @@ def cloudfoundry_tn(self, flavor):
...
+                ),
+            )
+        ]
+    )
+
+    # for kube-nodes and kube-systems
+    data['fvTenant']['children'][0]['fvAp']['children'][1]['fvAEPg']['children'].append(consumer_os_contract)
+    data['fvTenant']['children'][0]['fvAp']['children'][2]['fvAEPg']['children'].append(provider_os_contract)
+
+    # add new subject under contract
```
snaiksat: this comment should be:
add new contract and subject

provision/acc_provision/apic_provision.py
```diff
@@ -3415,7 +3414,223 @@ def cloudfoundry_tn(self, flavor):
...
+    misc1 = collections.OrderedDict(
```
snaiksat: instead of:
misc1

why not call it:
openshift-svc-catalog-contract

PR #38, opened 2019-03-25
Adding latest versions at various places for the upcoming release 4.1.1

gaurav-dalvi:


provision/acc_provision/versions.yaml
```diff
@@ -17,3 +17,9 @@ versions:
     aci_containers_controller_version: 1.9r43
     opflex_agent_version: 1.9r80
     openvswitch_version: 1.7r24
+  4.1.1:
```
snaiksat: So that we dont have to change this for 4.1.2 (and any subsequent releases), why not just keep this to 4.1?
Also may be we should just remove 1.6 and 1.7.

utils/python/edit_versions_yaml.py
```diff
@@ -33,7 +33,7 @@ def main():
     parser = optparse.OptionParser()
     parser.add_option("--branch-id",
                       dest="release_branch",
-                      help="one of them : 1.7, 1.8 or 1.9")
+                      help="one of them : 1.7, 1.8 or 1.9, 4.1.1")
```
snaiksat: help="1.9 or 4.1")

provision/acc_provision/versions.yaml
gaurav-dalvi: ok. I will make that change.

PR #46, opened 2019-03-30
Adding openshift-monitoring related ports, contracts, filters

gaurav-dalvi:


snaiksat:
I like this approach. I think we need to add kube-system providing kube-api contract in case of openshift.

gaurav-dalvi:
We need to know for which contract as well. For eg : this is the format:

```
aci_config:
        items:
          - name: openshift-svc-catalog
            range: [2379, 2379]
            etherT: ip
            prot: tcp
            stateful: "no"
            consumed: ["kube-system"]
            provided: ["kube-nodes"]
          - name: openshift-monitoring
            range: [9091, 9094]
            etherT: ip
            prot: tcp
            stateful: "no"
            provided: ["kube-system","kube-default","kube-nodes"]
            consumed: ["kube-system","kube-default"]
```
for eg: openshift-monitoring contract - kube-system, kube-default and kube-nodes are providing and kube-system and kube-default are consuming. this contract will open 9091 and 9094 tcp ports.

snaiksat:
Contract name is "kube-api". Currently "kube-system" is consuming it, needs to provide as well.

snaiksat:
It will be helpful if you can state in the commit message a little more details around what is going on, something along the lines of:

A new contract "openshift-monitoring” with the following details:
tcp: 9091, 9094
Provided by: kube-system, kube-default, kube-nodes
Consumed by: kube-system, kube-default

In addition "kube-system" will also now provide the “kube-api” contract (currently its only consuming).

Also the flavors definition in flavors.yaml is being updated to allow adding contract relationships (provided/consumed).

gaurav-dalvi:
Done @snaiksat . Updated commit message.

PR #62, opened 2019-04-12
Generalize APIC resources name according to the container ecosystem

apoorva11029:
Added option to add "app_profile" as part of the input file to acc-provision, defaults according to the flavor name: "openshift", "dockerucp", and kubernetes for the rest.

More info: noironetworks/support#845

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -247,6 +247,7 @@ def cidr_split(cidr):
 
 def config_adjust(args, config, prov_apic, no_random):
     system_id = config["aci_config"]["system_id"]
+    app_profile = config["aci_config"]["app_profile"]
     infra_vlan = config["net_config"]["infra_vlan"]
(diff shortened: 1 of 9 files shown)


puppet-py:
Since we want to breakout from our generalized app_profile name(currently kubernetes used for both K8s and Openshift deployment), we should set the default string on checking with the flavor. In this current PR, if Openshift user does not explicitly mention the app_profile we will land up creating "kubernetes" which defies our intent.
So I would suggest, create the default string(app_profile) based on the flavor passed

apoorva11029:
@puppet-py added the change

puppet-py:
We have one more flavor "cloudfoundry" :-), the app_profile corresponding to it should "cloud-foundry". Sorry missed out in pointing this to you earlier in the email-thread

apoorva11029:
@puppet-py we don't need to provide an option for setting the app_profile for cloudfoundry as it's handled separately in the code already

PR #64, opened 2019-05-31
Add refreshtime parameter (Optional) to acc-provision.

kishorejonnala:
To leverage APIC's custom subscription refresh time-out functionality,
an optional parameter "apic_refreshtime" is added to provision-config file.
When this parameter is set, Class/MO based subscriptions from ACC to APIC
will be refreshed based on the value set by this parameter.

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -386,6 +386,10 @@ def config_adjust(args, config, prov_apic, no_random):
             "configuration_version": token,
         }
     }
+    if config["aci_config"].get("apic_refreshtime"):  # APIC Subscription refresh timeout value
+        apic_refreshtime = config["aci_config"]["apic_refreshtime"]
+        adj_config["aci_config"]["apic_refreshtime"] = apic_refreshtime
+
     if config["net_config"].get("vip_subnet"):
         vip_subnet = cidr_split(config["net_config"]["vip_subnet"])
         adj_config["cf_config"]["app_vip_pool"] = [
@@ -422,6 +426,21 @@ def is_valid_mtu(xval):
     raise(Exception("Must be integer between %d and %d" % (xmin, xmax)))
 
 
+def is_valid_refreshtime(xval):
+    if xval is None:
+        # Not a required field.
+        return True
+    xmin = 0
+    xmax = (12 * 60 * 60)  # 12Hrs is the max suggested subscription refresh time for APIC
+    try:
+        x = int(xval)
+        if xmin <= x <= xmax:
+            return True
+    except ValueError:
+        pass
+    raise(Exception("Must be integer between %d and %d" % (xmin, xmax)))
+
+
 def config_validate(flavor_opts, config):
     def Raise(exception):
         raise exception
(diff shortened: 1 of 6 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -122,6 +122,7 @@ def config_default():
     default_config = {
         "aci_config": {
             "system_id": None,
+            "refreshtime": 0,  # APIC Subscription refresh interval; default to 0 to handle APIC older versions.
```
snaiksat: 1. To preserve existing behavior, we actually don't want to set a default here (because 0 is not default on the aci-container-controller side). The user would set to 0 explicitly if he wants the older (3.2x) behavior.
2. Can you call it: apic_refreshtime. I know this goes under the apic section and hence adding "apic_" prefix might seem redundant, but it still helps to more easily correlate with the configuration used in the aci-container-controller.

gaurav-dalvi:
Please update the main title and description. Looks like its rolled over from title to description.

provision/acc_provision/acc_provision.py
gaurav-dalvi: nit pick : refreshtime -> refresh_time

provision/acc_provision/acc_provision.py
```diff
@@ -422,6 +425,21 @@ def is_valid_mtu(xval):
     raise(Exception("Must be integer between %d and %d" % (xmin, xmax)))
 
 
+def is_valid_refreshtime(xval):
+    if xval is None:
+        # default 0 will be used
```
gaurav-dalvi: Add printf here 
https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/acc_provision.py#L71

provision/acc_provision/acc_provision.py
snaiksat: I believe not having the '_' was intentional to be consistent with the naming on the aci-container-controller side.

kishorejonnala: Yes, it was intentional to be consistent.

kishorejonnala: > 1. To preserve existing behavior, we actually don't want to set a default here (because 0 is not default on the aci-container-controller side). The user would set to 0 explicitly if he wants the older (3.2x) behavior.

So, you prefer to keep the default as 900 from acc side? I thought defaulting to 0 inturn results into APIC default timeout 90s and gracefully take care of lower versions. Any non-default will be specified by the user anyway. Let me know if you still want me to remove the default 0 setting, will take it out.

kishorejonnala:
> Please update the main title and description. Looks like its rolled over from title to description.

Fixed it!

provision/acc_provision/acc_provision.py
snaiksat: > So, you prefer to keep the default as 900 from acc side?
Yes.

> I thought defaulting to 0 inturn results into APIC default timeout 90s and gracefully take care of lower versions.
No. And that is because in 3.2.x we cannot send the refreshtime argument in the URI, the APIC will not ignore it, it will throw an error. So we have effectively two defaults on the acc side. One is for 4.x, and another one for 3.2.x. The one for 4.x is actually treated as default (i.e. 900), but if we want the 3.2x behavior, it can be achieved by setting to 0 (which in turn ends up not sending the refreshtime argument, and picking up the default on the APIC side). 

> Let me know if you still want me to remove the default 0 setting, will take it out.
Yes, and thanks! As a side-effect, I believe the existing UTs would not have to be updated either.

provision/testdata/with_refreshtime.apic.txt
```diff
@@ -0,0 +1,916 @@
+/api/mo/uni/infra/vlanns-[kube-pool]-static.json
```
snaiksat: I believe there is no refresh time being sent in anywhere in this apic configuration. As such, can we not use an existing output file for validation? If so, I would prefer doing that as opposed to creating a new one. Else for someone not familiar, it gives the impression that there is something more specific to the refresh time that you are trying to validate here and hence creating the new file (since the whole file is new, there is no diff with a previous version).

gaurav-dalvi:
Please squash all the commits into one

PR #78, opened 2019-08-01
SNAT operator changes

apoorva11029:
Add snat operator container to acc pod

### DIFF
--- a/provision/acc_provision/versions.yaml
+++ b/provision/acc_provision/versions.yaml
@@ -3,11 +3,13 @@ versions:
     cnideploy_version: 1.9r43
     aci_containers_host_version: 1.9r43
     aci_containers_controller_version: 1.9r43
+    snat_operator_version: snat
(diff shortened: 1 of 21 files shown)


provision/acc_provision/versions.yaml
```diff
@@ -3,11 +3,13 @@ versions:
     cnideploy_version: 1.9r43
     aci_containers_host_version: 1.9r43
     aci_containers_controller_version: 1.9r43
+    snat_operator_version: snat
     opflex_agent_version: 1.9r80
     openvswitch_version: 1.7r24
   4.1:
     aci_containers_controller_version: 4.1.1.5.r19
     aci_containers_host_version: 4.1.1.5.r19
+    snat_operator_version: snat
```
snaiksat: Would have been preferable to use "latest" but since we will change this before release, its not a big deal. Also at that time we should move this to the version 4.2 (and remove from version 4.1).

PR #82, opened 2019-08-19
Change OVS memory limit to 10G

kishorejonnala:
* Also make it configurable

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -171,6 +171,7 @@ def config_default():
             "image_pull_policy": "Always",
             "kubectl": "kubectl",
             "system_namespace": "kube-system",
+            "ovs_memory_limit": "10Gi",
             "snat_operator": {
                 "name": "snat-operator",
                 "watch_namespace": "default",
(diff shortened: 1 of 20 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -180,6 +180,7 @@ def config_default():
         },
         "registry": {
             "image_prefix": "noiro",
+            "memory_limit": "10Gi",
```
snaiksat: Thanks for the quick turnaround on this. This section is probably not the right place to put this config (since it refers to the docker registry). Perhaps the kube_config section above is a better place. Also, if you can rename memory_limit to ovs_memory_limit, that might be more informative to the user. (I know changes to the name would imply changing it in all places in the tests, but hopefully you can run a script to do it.

kishorejonnala: Makes sense. Fixed.

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -61,3 +61,6 @@ net_config:
 registry:
   image_prefix: noiro                   # e.g: registry.example.com/noiro
   # image_pull_secret: secret_name      # (if needed)
+
+kube_config:
+  ovs_memory_limit: "10Gi"       #e.g:  memory: "10Gi"
```
snaiksat: Hate to nit pick, but we seem to be following the convention of commenting out any configuration which is optional for the user to set, and leave only that configuration uncommented which the user absolutely needs to set/override. So my suggestion would be to say:
  # ovs_memory_limit: "10Gi"       # Override if needed, default is 10Gi

PR #105, opened 2019-09-07
Configure external pod connectivity enabled

apoorva11029:
Enable only when "allow_pods_kube_api_access" is true


snaiksat:
Thanks for fixing this!

PR #109, opened 2019-09-09
Enable PBR Tracking by default

kishorejonnala:
- Set IPSLA interval to 5, this enables PBR tracking
- User can disable PBR tracking by setting the interval to 0

This is mainly done for SNAT feature as all compute-nodes are added as service endpoints and it is important to make sure the reachability of these nodes.


snaiksat:
Please backport.

snaiksat:
Actually lets hold off on back porting this to release branch. I think we discussed to not turn this ON by default for the I release. Based on feedback we will turn it ON in the upcoming release.

PR #110, opened 2019-09-10
Add configmap for snat portRange

apoorva11029:
To complement https://github.com/noironetworks/snat-operator/pull/26

Add a new configMap called snat-operator-config in namespace aci-containers-system to configure port ranges and ports assigned per node during SNAT

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -176,6 +176,11 @@ def config_default():
                 "name": "snat-operator",
                 "watch_namespace": "",
                 "globalinfo_name": "snatglobalinfo",
+                "port_range": {
+                    "start": 5000,
+                    "end": 65000,
+                    "ports_per_node": 3000,
+                },
             },
         },
         "registry": {
--- a/provision/acc_provision/templates/aci-containers.yaml
+++ b/provision/acc_provision/templates/aci-containers.yaml
@@ -282,6 +282,19 @@ data:
     }
 ---
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: snat-operator-config
+  namespace: aci-containers-system
+  labels:
+    aci-containers-config-version: "{{ config.registry.configuration_version }}"
+    network-plugin: aci-containers
+data:
+    "start": "{{config.kube_config.snat_operator.port_range.start}}"
+    "end": "{{config.kube_config.snat_operator.port_range.end}}"
+    "ports-per-node": "{{config.kube_config.snat_operator.port_range.ports_per_node}}"
+---
+apiVersion: v1
 kind: Secret
 metadata:
   name: aci-user-cert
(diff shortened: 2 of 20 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -176,6 +176,11 @@ def config_default():
                 "name": "snat-operator",
                 "watch_namespace": "",
                 "globalinfo_name": "snatglobalinfo",
+                "portRange": {
+                    "start": 5000,
+                    "end": 65000,
+                    "portsPerNode": 3000,
```
snaiksat: Hate to nit pick, but I think we should be consistent with the naming convention. Can we change portsPerNode to ports_per_node?
Also, can we change portRange to snat_port_range?

provision/testdata/base_case.kube.yaml
```diff
@@ -248,6 +248,19 @@ data:
...
+metadata:
+  name: snat-operator-config
+  namespace: aci-containers-system
+  labels:
+    aci-containers-config-version: "dummy"
+    network-plugin: aci-containers
+data:
+    "start": "5000"
+    "end": "65000"
+    "portsPerNode": "3000"
```
snaiksat: Why are these strings, should they not be integers?

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -282,6 +282,19 @@ data:
...
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: snat-operator-config
+  namespace: aci-containers-system
+  labels:
+    aci-containers-config-version: "{{ config.registry.configuration_version }}"
+    network-plugin: aci-containers
+data:
+    "start": "{{config.kube_config.snat_operator.portRange.start}}"
```
snaiksat: Should this be: config.kube_config.snat_operator.portRange.start|json
to parse as integer (here and below)?

provision/testdata/base_case.kube.yaml
apoorva11029: K8s doesn't allow you to add integer values in ConfigMap in this format. Either we use a JSON file like we have used in the configMap, which we moved away from because "getting" it directly from k8s client was giving a long string. 
ref: https://medium.com/@ismailyenigul/kubernetes-configmap-and-cannot-convert-int64-to-string-error-ebf48fa1cd28

provision/acc_provision/templates/aci-containers.yaml
apoorva11029: Since we are not parsing it as an integer

provision/acc_provision/acc_provision.py
apoorva11029: Yes sure! I just reused the names I used in snat-operator code but here it makes more sense to use underscores.

PR #116, opened 2019-09-13
Config for max nodes in snat svc graph; default-32

apoorva11029:
Set option to configure the maximum number of nodes allowed per service graph

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -184,6 +184,7 @@ def config_default():
                     "ports_per_node": 3000,
                 },
             },
+            "max_nodes_svc_graph": 32
         },
         "registry": {
             "image_prefix": "noiro",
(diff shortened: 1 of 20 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -184,6 +184,7 @@ def config_default():
                     "ports_per_node": 3000,
                 },
             },
+            "max_nodes_svc_graph": 24
```
snaiksat: this should be 32, here and below

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -222,6 +222,7 @@ data:
             "policy-space": {{config.kube_config.default_endpoint_group.tenant|json}},
             "name": {{(config.kube_config.default_endpoint_group.app_profile ~ "|" ~ config.kube_config.default_endpoint_group.group)|json}}
         },
+        "max-nodes-svc-graph": {{ config.kube_config.max_nodes_svc_graph|json }},
```
snaiksat: Can we also validate that its > 0 and <= 64

apoorva11029: Added!

provision/acc_provision/acc_provision.py
apoorva11029: Thanks, changed it to 32.

PR #121, opened 2019-09-18
Option to enable PBR-Tracking for non-snat services

kishorejonnala:
When IPSLA is set to non-zero value, it takes effect only for SNAT services and the user is required to enable "pbr_tracking_non_snat" explicitly for non-snat services.

This is for not taking a chance of breaking any existing functionality during ACC upgrade.


snaiksat:
It will be good to have a test with this set to True (to exercise the non-default value). Can be added as a follow up.

PR #125, opened 2019-09-18
Add snat namespace field to host config

apoorva11029:
Complementing https://github.com/noironetworks/aci-containers/pull/339
Set to "aci-containers-sytem" by default. Can be overridden by changing kube_config/snat_operator/snat_namespace field

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -185,6 +185,7 @@ def config_default():
                     "end": 65000,
                     "ports_per_node": 3000,
                 },
+                "snat_namespace": "aci-containers-system",
             },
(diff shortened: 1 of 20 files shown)


provision/acc_provision/acc_provision.py
```diff
@@ -185,6 +185,7 @@ def config_default():
                     "end": 65000,
                     "ports_per_node": 3000,
                 },
+                "snat_namespace": "aci-containers-system",
```
snaiksat: Per discussion, would prefer to not call this snat_namespace as its more generic, but perhaps we can add a different conf as follow up and deprecate this conf at that time.

PR #137, opened 2019-09-24
Unprovisioning code for tagAnnotation

apoorva11029:
For APIC version >= 3.2, our APIC objects can now be deleted using acc-provision -d as well

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -3,7 +3,7 @@ from __future__ import print_function, unicode_literals
 import collections
 import json
 import sys
-
+import re
 import requests
 import urllib3
 import ipaddress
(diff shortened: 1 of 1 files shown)


provision/acc_provision/apic_provision.py
```diff
@@ -221,28 +237,51 @@ def valid_tagged_resource(self, tag, system_id, tenant):
...
-                    mo = mo_dict[mo_key]
-                    mo_dn = mo["attributes"]["dn"]
-                    mos[mo_dn] = True
-                    dbg("    - %s" % mo_dn)
+                    tag_name = tag_mo["tagAnnotation"]["attributes"]["value"]
+                    if self.valid_tagged_resource(tag_name, system_id, tenant):
+                        dbg("Deleting tag: %s" % tag_name)
+                        dn = tag_mo["tagAnnotation"]["attributes"]["dn"]
+                        reg = re.search('(.*)(/annotationKey.*)', dn)
+                        dn_name = reg.group(1)
```
kishorejonnala: Call this as parent_dn

apoorva11029: Done!

provision/acc_provision/apic_provision.py
```diff
@@ -221,28 +237,51 @@ def valid_tagged_resource(self, tag, system_id, tenant):
     def clean_tagged_resources(self, system_id, tenant):
         tags = collections.OrderedDict([])
         tags_path = "/api/node/mo/uni/tn-%s.json" % (tenant,)
-        tags_path += "?query-target=subtree&target-subtree-class=tagInst"
+        apic_version = self.get_apic_version()
```
snaiksat: Should we always do the check for tagged resources as well? This might perhaps manifest in the case where upgrade has not been done and previous install had tagged resources?

kishorejonnala: @snaiksat I agree with you. It's better to do tagInst always regard less of the version

kishorejonnala: @snaiksat  Are you referring to check both TagInst, TagAnnotation resources regardless of APIC version? I don't see a problem in doing that.

@apoorva11029  You can take out GetVersion code if you agree.

PR #141, opened 2019-09-26
Upgrade to release 4.2

apoorva11029:


snaiksat:
Perhaps not critical but this can be updated:
https://github.com/noironetworks/acc-provision/blob/master/utils/python/edit_versions_yaml.py#L37

PR #143, opened 2019-10-03
Remove default flavor option from acc-provision

apoorva11029:
Return an error if no flavor provided.


snaiksat:
As a follow up, it might be helpful to have a negative test case also such that you check the stdout/stderr when no flavor is provided.

PR #150, opened 2019-10-31
Overlay changes and add eks flavor

apoorva11029:
Introduce host-config variables called ep_registry and opflex-mode for overlay mode. opflex-mode is set to "overlay" for overlay mode.

Add a new flavor for EKS.


provision/acc_provision/acc_provision.py
```diff
@@ -175,6 +176,7 @@ def config_default():
             "system_namespace": "kube-system",
             "ovs_memory_limit": "1Gi",
             "reboot_opflex_with_ovs": "true",
+            "ssl_mode": "disabled",
```
snaiksat: Will this disable the ssl_mode by default? If so, its not backward compatible with the on-prem solution and we can't do that.

apoorva11029: Sorry, I meant to change this. Reverting.

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -692,6 +727,10 @@ spec:
           env:
             - name: REBOOT_WITH_OVS
               value: "{{ config.kube_config.reboot_opflex_with_ovs }}"
+            {% if config.kube_config.run_gbp_container %}
+            - name: SSL_MODE
+              value: {{ config.kube_config.ssl_mode }}
```
snaiksat: This might be the place to disable it.

apoorva11029: Added.

provision/testdata/flavor_localhost.inp.yaml
```diff
@@ -28,13 +28,18 @@ aci_config:
 net_config:
   node_subnet: 1.100.202.1/24              # Subnet to use for nodes
   pod_subnet: 10.2.56.1/21 # Subnet to use for Kubernetes Pods
+  gbp_pod_subnet: 10.3.56.1/21
```
jojimt: Please derive the gbp_pod_subnet in gw/subnet_length form using the cni gateway information and the pod subnet network length, as discussed.

provision/acc_provision/acc_provision.py
```diff
@@ -304,6 +308,7 @@ def config_adjust(args, config, prov_apic, no_random):
         },
         "net_config": {
             "infra_vlan": infra_vlan,
+            "gbp_pod_subnet": "%s/%s" % (cidr_split(pod_subnet)[2], cidr_split(pod_subnet)[4]),
```
snaiksat: So this will always get set, right? Should be okay as long we don't have conditional overlay-only specific logic derived from it.

apoorva11029: This always get set, but we don't use this as a conditional for overlay solution

PR #151, opened 2019-10-31
Remove namespace label for k8s flavors

apoorva11029:
Fix for https://cdetsng.cisco.com/webui/#view=CSCvr88382
Remove the label network-plugin: aci-containers from aci-containers-system created for non-openshift flavors, as it was breaking acikubectl cluster-report functionality

### DIFF
--- a/provision/acc_provision/templates/aci-containers.yaml
+++ b/provision/acc_provision/templates/aci-containers.yaml
@@ -17,7 +17,6 @@ metadata:
   name: aci-containers-system
   labels:
     aci-containers-config-version: "{{ config.registry.configuration_version }}"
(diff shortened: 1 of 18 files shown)


snaiksat:
Should there be a test case which validates that the label is added in the case of OpenShift flavor?

apoorva11029:
> Should there be a test case which validates that the label is added in the case of OpenShift flavor?
@snaiksat 
This is the test case to validate the label is added in Openshift: https://github.com/noironetworks/acc-provision/blob/5bdc0769b283d8b8bc80c12c1d909626b73751a9/provision/testdata/flavor_openshift.kube.yaml#L7

PR #167, opened 2019-12-12
Ability to add a cluster to a pre-existing tenant

apoorva11029:
1. Introduce a new (dictionary)field called "tenant" in acc-provision input file to provision cluster in an existing tenant, like this:
aci_config:
  tenant:
    name: existing_tenant

2. The cluster will be provisioned with a new naming convention, most apic objects that were named kube-, will be now called aci-containers-<system_id>- to avoid ambiguity.

3. For supporting existing clusters with the kube convention, a boolean field is introduced inside aci_config like this:
aci_config:
  use_legacy_kube_naming_convention: True
However, pre-existing tenant and use_kube_naming_convention can't be set at the same time. 

4. Code changes to run all the ACI CNI pods in the namespace aci-containers-system instead of kube-system.

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -136,6 +136,11 @@ def config_default():
     default_config = {
         "aci_config": {
             "system_id": None,
+            "tenant": {
+                "name": None,
+            },
+            "use_pre_existing_tenant": False,
+            "use_legacy_kube_naming_convention": False,
             "vrf": {
                 "name": None,
                 "tenant": None,
@@ -184,7 +189,7 @@ def config_default():
             "use_cluster_role": True,
             "image_pull_policy": "Always",
             "kubectl": "kubectl",
-            "system_namespace": "kube-system",
+            "system_namespace": "aci-containers-system",
             "ovs_memory_limit": "1Gi",
             "reboot_opflex_with_ovs": "true",
             "snat_operator": {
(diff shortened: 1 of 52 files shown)


snaiksat:
I believe the "kube-" convention was used earlier to distinguish kubernetes related (and acc-provision created) objects from any others created in the same tenant. Would it be helpful to preserve the "kube-" prefix?

apoorva11029:
> I believe the "kube-" convention was used earlier to distinguish kubernetes related (and acc-provision created) objects from any others created in the same tenant. Would it be helpful to preserve the "kube-" prefix?

@snaiksat preserved it by adding "aci-containers" prefix to all the provisioned object names.

snaiksat:
nit comment: in the PR description, can you change "name: old_tenant" to name: existing_tenant"?

apoorva11029:
> nit comment: in the PR description, can you change "name: old_tenant" to name: existing_tenant"?

@snaiksat updated the PR description and commit message

provision/acc_provision/acc_provision.py
```diff
@@ -743,6 +772,20 @@ def config_validate_preexisting(config, prov_apic):
...
+            # naming convention. This is a fallback in case the user
+            # forgets to add the field to indicate an existing legacy
+            # cluster.
+            # if kube_ap:
+            #     config["aci_config"]["use_legacy_kube_naming_convention"] = True
+            #     if config["aci_config"]["vmm_domain"]["type"] == "OpenShift":
+            #         config["kube_config"]["system_namespace"] = "aci-containers-system"
+            #     else:
+            #         config["kube_config"]["system_namespace"] = "kube-system"
+
```
snaiksat: do we need this? (btw, minor typo: "lagacy")

apoorva11029: Commented out for now so that it can be added to the code later.

provision/acc_provision/acc_provision.py
```diff
@@ -487,6 +512,10 @@ def config_adjust(args, config, prov_apic, no_random):
             adj_config['cf_config']['default_endpoint_group']['app_profile'],
             adj_config['cf_config']['node_epg']))
 
+    if config["aci_config"]["vmm_domain"]["type"] == "Kubernetes":
```
snaiksat: Per discussion, it seems this check should be removed.

apoorva11029:
Done!

provision/acc_provision/acc_provision.py
apoorva11029: Removed.

snaiksat:
One general comment, as a follow up, we should refactor the test cases such that the new naming convention is what gets tested across the tests (and the use_prexisting_tenant flag is set only for testing the pre-existing case specifically).

PR #169, opened 2020-01-07
Istio install:

kishorejonnala:
Changes to auto install Istio by default along with ACI-CNI deploy.
=> Uses Istio's standalone operator model
=> Results in the creation of istio-system namespace
=> This namespace will be annotated by ACC to map with an EPG named kube-istio
=> acc-provision will create a new EPG called kube-istio on APIC
=> acc-provision also creates a new contract "istio" which will be provided by kube-istio and consumed by kube-default
=> Contract "istio" is created with "istio-subject" which will have filter entries to enable istio contol-plane ports.
=> kube-istio consumes "kube-api" contract


provision/acc_provision/apic_provision.py
```diff
@@ -2830,6 +3027,217 @@ def kube_tn(self, flavor):
...
+                                                                                                    "etherT",
+                                                                                                    "ip",
+                                                                                                ),
+                                                                                                (
+                                                                                                    "prot",
+                                                                                                    "tcp",
+                                                                                                ),
+                                                                                                (
+                                                                                                    "sFromPort",
+                                                                                                    "9080",
```
apoorva11029: Could be a good idea to drive these values from input file eventually.

kishorejonnala: Sure, as discussed - this may not be trivial; will sure give it a try in a later commit.

PR #172, opened 2020-01-18
Mount hostvar into opflex-server for grpc config

jojimt:


jojimt:
@snaiksat @apoorva11029 can one of you please review this? This change is needed for passing the grpc info to the opflex-server,

PR #173, opened 2020-01-22
Changes for new snat design

apoorva11029:
Remove snatlocalinfo and move the snat-operator container logic into the aci-containers-controller container.

### DIFF
--- a/provision/acc_provision/templates/aci-containers.yaml
+++ b/provision/acc_provision/templates/aci-containers.yaml
@@ -94,20 +94,6 @@ spec:
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: snatlocalinfos.aci.snat
-spec:
-  group: aci.snat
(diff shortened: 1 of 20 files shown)


snaiksat:
The following were created for the "snat-operator", are the still needed as separate entities, or can they be collapsed into the existing config-maps, etc. (we can do these changes in one or more followup patches, and I imagine coordinated changes might be needed on the acc side as well):
snat-operator-config
aci-snat-namespace

apoorva11029:
> The following were created for the "snat-operator", are the still needed as separate entities, or can they be collapsed into the existing config-maps, etc. (we can do these changes in one or more followup patches, and I imagine coordinated changes might be needed on the acc side as well):
> snat-operator-config
> aci-snat-namespace

Good point. Let's do it in the next iteration

snaiksat:
Also, for whichever CRDs we are changing the schema, we might need to update the version. Let's investigate as a follow up as to what exactly needs to be done.

PR #182, opened 2020-01-28
Fix cidr, rbac

jojimt:


apoorva11029:
Looks good, can we add a UT to test the deployment yaml for EKS here: https://github.com/noironetworks/acc-provision/blob/eebd26ff580498e37d85070db16417c309cd58ed/provision/acc_provision/test_main.py#L221?

jojimt:
We need to separate out the cAPIC part before we can add the test.

PR #200, opened 2020-02-25
AciIstioOperator: Moving upstream istio-operator consumption to ACC

kishorejonnala:
- With the introduction of AciIstioOperator, ACC will directly consume upstream istio-operator and trigger istio install.
- This change will add AciIstioOperator CRD to deployment yaml and remove upstream istio-operator/CRD from deployment file.
- Also, a new container "aci-istio-operator" is added to aci-containers-controller pod.
- Modified ClusterRole attributes as necessary.


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -659,6 +428,34 @@ rules:
...
+  - istiocontrolplanes
+  verbs:
+  - '*'
+- apiGroups:
+  - "aci.istio"
+  resources:
+  - aciistiooperators
+  - aciistiooperator
+  verbs:
+  - '*'
```
snaiksat: revisit  the * in a follow up PR

kishorejonnala: Sure, will do.

PR #201, opened 2020-02-25
Ability to generate tar.gz of deployment files(for OpenShift)

apoorva11029:
Use -z argument in acc-provision to export a tar.gz file with all deployment and operator yamls saved as separate tar files

If acc-provision is run with just -o <filename>, automatically save the tar with the name filename.tar.gz in the working directory
If acc-provision is run with -o <filename> -z <path-to-tarname.tar.gz>, save tarname.tar.gz to the path and deployment yaml to filename
If acc-provision is run with just -t <path-to-tarname.tar.gz>, save the tar to path and print the deployment yaml to stdout

The files in the tar have naming convention: cluster-network-<file-number>-<k8sobject_kind>-<k8sobject-name>.yaml
The file number is starting from 4 because the first 3 are needed by OpenShift.

Fixed https://github.com/noironetworks/support/issues/1076
Unit-testing: tweaked base_case test to validate generated tar

Also modified the aci-operators.yaml template to remove redundant yamls


provision/acc_provision/acc_provision.py
```diff
@@ -959,48 +986,73 @@ def generate_kube_yaml(config, output, operator_output, operator_cr_output):
...
-            op_cr_template.stream(config=config).dump(operator_cr_output)
+        # The next few files are to generate tar file with each
+        # containers and operator yaml in separate file. This is needed
+        # by OpenShift >= 4.3. If tar_path is provided(-z), we save the tar
+        # with that filename, else we use the provided containers
+        # deployment filepath. If neither is provided, we don't generate
+        # the tar.
+        if tar_path == "-":
+            tar_path = "/dev/null"
+        if tar_path != "/dev/null":
```
snaiksat: perhaps use else here

provision/acc_provision/acc_provision.py
```diff
@@ -946,9 +947,35 @@ def get_jinja_template(file):
...
+    filenames = []
+
+    # Function to construct filenames for each yaml
+    def gen_file_list(docs, start, filenames):
+        for doc in docs:
+            filename = "cluster-network-" + str(start).zfill(2) + "-" + doc['kind'] + "-" + doc['metadata']['name'] + ".yaml"
+            filenames.append(os.path.basename(filename))
+            with open(filename, 'w') as outfile:
+                yaml.safe_dump(doc, outfile, default_flow_style=False, encoding="utf-8")
+            start += 1
```
snaiksat: may be call this "counter"

provision/acc_provision/acc_provision.py
```diff
@@ -946,9 +947,35 @@ def get_jinja_template(file):
...
+        return start
+
+    file_start = gen_file_list(cont_docs, file_start, filenames)
+    gen_file_list(oper_docs, file_start, filenames)
+
+    # Create tar for the parsed files and delete the files too
+    tar = tarfile.open(tar_path, "w:gz", encoding="utf-8")
+    for name in filenames:
+        tar.add(name)
+        os.remove(name)
```
snaiksat: Perhaps create a new dir, and the tar all files in that dir to simplify some of the handling above (this can be a follow up PR).

provision/acc_provision/test_main.py
```diff
@@ -481,40 +482,66 @@ def get_args(**overrides):
...
+def compare_yaml(expectedyaml, output, debug, generated):
+    if expectedyaml is not None:
+        if debug:
+            shutil.copyfile(output.name, generated)
+        with open(expectedyaml, "r") as expected:
+            assert output.read() == expected.read()
+
+
+def tarsum(input_file, hash):
+    """
```
snaiksat: Maybe we can just use: hashlib.md5(open('<filename>','rb').read()).hexdigest()

PR #220, opened 2020-03-13
Fixes from integration with IPI

jojimt:
Incorporate changes from new naming, plus some cAPIC model related changes.


provision/acc_provision/acc_provision.py
```diff
@@ -1491,24 +1511,108 @@ def addMiscConfig(config):
...
+        m_cidr = config["net_config"]["machine_cidr"]
+        b_subnet = config["net_config"]["bootstrap_subnet"]
+        n_subnet = config["net_config"]["node_subnet"]
+        p_subnet = config["net_config"]["pod_subnet"].replace(".1/", ".0/")
+        region = config["aci_config"]["vrf"]["region"]
+        boot_subnetID = getSubnetID(b_subnet)
+        node_subnetID = getSubnetID(n_subnet)
+        print("\nOpenshift Info")
+        print("----------------")
+        print("networking:\n  clusterNetwork:\n  - cidr: {}\n    hostPrefix: 23\n  machineCIDR: {}\n  networkType: CiscoAci\n  serviceNetwork:\n  - 172.30.0.0/16\nplatform:\n  aws:\n    region: {}\n    subnets:\n    - {}\n    - {}".format(p_subnet, m_cidr, region, boot_subnetID, node_subnetID))
```
apoorva11029, line 1615: Do we need this print statement? Maybe in debug mode?

provision/acc_provision/flavors.yaml
```diff
@@ -381,6 +383,89 @@ flavors:
...
+            consumed: ["inet-ext"]
+          - name: allow-cidr
+            underlay: True
+            filter: match-oshift
+            provided: ["ul-boot", "ul-nodes"]
+            consumed: ["cidr-ext"]
+          - name: allow-kube
+            underlay: False
+            filter: match-all
+            provided: ["aci-containers-default", "aci-containers-nodes", "aci-containers-system"]
```
apoorva11029, line 408: The "aci-containers-" prefix will work only when the "use_legacy_kube_naming_convention" field is not set. 
Is that an assumption here, not never have the older naming convention?

For OpenShift flavor, we use this: https://github.com/noironetworks/acc-provision/blob/5fde656c560c12c6a311e1d1d3379a5230e49bd3/provision/acc_provision/flavors.yaml#L121
And add the prefix based on the naming conventon.

provision/acc_provision/acc_provision.py
jojimt: This needs to be copied into the install_config.yaml, so we need this printed until we have a way to automate that edit.

provision/acc_provision/flavors.yaml
jojimt: Because this is a new flavor, my thought is we don't need to worry about legacy.

apoorva11029: Can we add a check that doesn't allow the field to be set in this flavor then?

jojimt: sure.

PR #243, opened 2020-04-02
Drop Log provisioning

shastrinator:
Config knob to disable drop-log events
Rest of the config is static and part of host-agent


provision/acc_provision/acc_provision.py
```diff
@@ -238,6 +238,9 @@ def config_default():
             "hostagent_log_level": "info",
             "opflexagent_log_level": "info",
         },
+        "drop_log_config": {
+            "enable": True
+        },
```
kishorejonnala, line 243: This is initializing part. You will have to read the configured value - look at config_adjust()
Ex: https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/acc_provision.py#L504

shastrinator: I don't have additional config, based on the knob, unlike istio, so I don't need to do anything more with config_adjust. I added the test case like you mentioned. Thanks for reviewing!

PR #255, opened 2020-04-30
Adding RBAC "create" to provide create access to aci operator for aciontainersoperator objeect

abhis2112:


provision/acc_provision/templates/aci-operators.yaml
```diff
@@ -67,10 +67,7 @@ rules:
   - acicontainersoperators/status
   - acicontainersoperators/finalizers
   verbs:
-  - get
-  - list
-  - update
-  - watch
+  - '*'
```
snaiksat, line 70: I would have still preferred that we enumerate all the verbs (even if its equivalent to *). We can discuss offline and consider as a follow up.

PR #288, opened 2020-07-09
Support for OpenShift 4.4 on ESX

apoorva11029:
Following changes are made from the existing OpenShift 4.3 flavor:
1. Create a "<systemid>_vlan_<kubeapi_vlan>" port group with vlan tag of kubeapi_vlan along with the current trunk port group that is being created
2. Do not create the 99-* manifests that are being created now as part of the tar file
3. Create a additional json file "apic.json" in the tar file with fields like network info and VLAN values- used to create the required install.config file for openshift installer.

Add a new testcase for this scenario called test_flavor_openshift_44_esx


provision/acc_provision/apic_provision.py
```diff
@@ -1933,6 +1934,136 @@ def nested_dom(self):
         self.annotateApicObjects(data)
         return path, data
 
+    def nested_dom_second_portgroup(self):
```
snaiksat, line 1937: As a follow up we can perhaps consider refactoring the vmmUsrCustomAggr into one method so that we can call it from two different places and avoid some repetition of code.

PR #298, opened 2020-07-17
Normalize pod and node CIDRs for ESX flavor

apoorva11029:


snaiksat:
Is it possible to add test case for just the normalize function? If it will take time, we can merge this and add as follow up.

provision/acc_provision/acc_provision.py
```diff
@@ -318,6 +318,18 @@ def cidr_split(cidr):
...
 
+def normalize_cidr(cidr):
+    # To convert CIDR network ending with .1 to .0. For eg, convert 10.0.0.1/16 to 10.0.0.0/16
+    rtr, mask = cidr.split('/')
+    ip = ipaddress.ip_address(rtr)
+    if ip.version == 4:
+        n = ipaddress.IPv4Network(cidr, strict=False)
+    else:
+        n = ipaddress.IPv6Network(cidr, strict=False)
+    normalized_cidr = "%s/%s" % (str(n[0]), mask)
```
snaiksat, line 329: In both these cases, would it just be preferable to do:

 n = ipaddress.IPv<4/6>Network(cidr, strict=False)
 str(n)

and not actually split, index, etc?

apoorva11029: Thanks for the suggestion, changed!

PR #299, opened 2020-07-18
Changes for openshift-4.4-openstack flavor

apoorva11029:
Similar to 4.3, just without kube_api_vlan or node bridge domain, and open ports 2380 and 9978 for etcd
A new option in aci_config called "disable_node_bd_creation" which causes node_bd to be created only if the option is set to False(True by default).
This option also takes care of not associating the node EPG to any BD in this case.


provision/testdata/list_flavors.stdout.txt
```diff
@@ -1,4 +1,5 @@
 INFO: Available configuration flavors:
+INFO: openshift-4.4-openstack:	Red Hat OpenShift Container Platform 4.4 on OpenStack
 INFO: openshift-4.4-esx:	Openshift 4.4 ESX
 INFO: openshift-4.4:	Red Hat OpenShift Container Platform 4.4
```
snaiksat, line 4: I am not sure that we will use this flavor anymore:
INFO: openshift-4.4:	Red Hat OpenShift Container Platform 4.4

So perhaps hide it for now.

provision/acc_provision/flavors.yaml
```diff
@@ -81,10 +85,110 @@ flavors:
...
+            prot: tcp
+            stateful: "no"
+            provided: ["system","default","nodes"]
+            consumed: ["system","default"]
+        kube_api_entries:
+          - name: etcd1
+            range: [2379, 2379]
+            etherT: ip
+            prot: tcp
+            stateful: "no"
```
snaiksat, line 147: I think we need to add 2380 here (so perhaps a range 2379-2380). Can you double check with abhishek?

apoorva11029: @abhis2112 can you please confirm?

provision/testdata/list_flavors.stdout.txt
apoorva11029: Done!

provision/acc_provision/flavors.yaml
abhis2112: @apoorva11029 yes, now we need for 2379, 2380, 9979, 9978
So the new ports are 2380 and 9978

apoorva11029: Added, thanks!

PR #304, opened 2020-07-24
Adding flavor file for cluster network DNS for Openshift 4.4

abhis2112:
Adding the flavor template file for cluster-network-22-DaemonSet-aci-containers-host.yaml for both OSP-4.4/4.3


abhis2112:
Merged in different PR

PR #305, opened 2020-07-24
acc provision template side changes for qospolicy

niteshcha:


niteshcha:
@tomflynn @snaiksat  please review this!

jayaramsatya:
Hi Nitesh, this requires changes in all other flavors of  yaml files. 

We have a script for that to make changes automatically by taking your template file as input.

clone this repository https://github.com/noironetworks/support.git 

and run script support/acc-scripts/acc-provision/fix-testdata.sh on top of your changes.

snaiksat:
@niteshcha could you kindly rebase your patch?

provision/Makefile
```diff
@@ -11,7 +11,7 @@ dist: test
 
 test:
 	flake8 --ignore E501,E731,E741,W504 acc_provision
-	python -m pytest acc_provision
+	python -m pytest -vv acc_provision
```
snaiksat, line 14: you probably used this for your debugging, can you please remove it?

niteshcha: done

apoorva11029:
base_case_qos.inp.yaml looks the same as base_case.inp.yaml. If that's the case, we don't need to add another testcase here

apoorva11029:
And do we need the qos template changes for all flavors?

niteshcha:
> base_case_qos.inp.yaml looks the same as base_case.inp.yaml. If that's the case, we don't need to add another testcase here

done

niteshcha:
> base_case_qos.inp.yaml looks the same as base_case.inp.yaml. If that's the case, we don't need to add another testcase here



> And do we need the qos template changes for all flavors?

yes

PR #344, opened 2020-08-12
Changes to provision for aks

jojimt:
- added a new flavor for aks
- added tests


provision/acc_provision/flavors.yaml
```diff
@@ -669,11 +669,63 @@ flavors:
     status: Experimental
     hidden: False
     order: 17
+  aks:
+    desc:         Azure Kubernetes Service
```
apoorva11029, line 673: nit: there appears to be a tab instead of space between desc: and Azure

provision/acc_provision/flavors.yaml
```diff
@@ -669,11 +669,63 @@ flavors:
...
+        filters:
+            - name: match-all
+              entries:
+                - name: ent1
+    status: Experimental
+    hidden: False
+    order: 18
   cloud:
-    desc:         Kubernetes Cloud
+    desc:         Openshift IPI/AWS
```
apoorva11029, line 723: same

PR #345, opened 2020-08-12
Disable node subnet creation for Openshift on Openstack flavor

apoorva11029:
Earlier the node bd wasn't being created, now we create it without the subnet

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -177,7 +177,7 @@ def config_default():
                 "activeFlowTimeOut": None,
             },
             "kube_default_provide_kube_api": False,
-            "disable_node_bd_creation": False,
+            "disable_node_subnet_creation": False,
             "preexisting_kube_bd": None,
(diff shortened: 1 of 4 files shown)


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -312,7 +312,9 @@ data:
         "service-vlan": {{ config.net_config.service_vlan|json }},
         "kubeapi-vlan": {{ config.net_config.kubeapi_vlan|json }},
         "pod-subnet": {{ config.net_config.pod_subnet|json }},
+        {% if not config.aci_config.disable_node_subnet_creation %}
         "node-subnet": {{ config.net_config.node_subnet|json }},
+        {% endif %}
```
snaiksat, line 317: We need to populate this regardless of whether we create the ACI subnet or not.

apoorva11029: Added back!

PR #348, opened 2020-08-15
Fix nodebd name when disable_node_bd_creation is set

apoorva11029:


provision/fix-testdata.sh
```diff
@@ -0,0 +1,270 @@
+rm -rf /tmp/generated*
```
snaiksat, line 1: we can remove this file

PR #358, opened 2020-08-31
Update vmmDomP/CtrlrP Mode for Rancher flavor

shastrinator:

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -460,10 +460,11 @@ class ApicKubeConfig(object):
         update(data, self.vdom_pool())
         update(data, self.mcast_pool())
         update(data, self.phys_dom())
-        update(data, self.kube_dom())
+        update(data, self.kube_dom(apic_version))
         update(data, self.nested_dom())
         update(data, self.associate_aep())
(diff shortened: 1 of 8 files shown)


tomflynn (dismissed):

shastrinator:
Updated plugin name to lowercase as per upstream comments.

tomflynn (approved):

provision/acc_provision/apic_provision.py
```diff
@@ -942,6 +944,8 @@ def kube_dom(self):
         elif vmm_type == "CloudFoundry":
             mode = "cf"
             scope = "cloudfoundry"
+        elif apic_version >= 5.1 and cluster_provider == "Rancher":
```
kishorejonnala, line 947: Why do we reply on apic_version to set the mode? Can we not just call it out explicitly?

tomflynn: rancher 'mode' was added in 5.1. so the post will fail with earlier versions. Rather than rejecting rancher usage with older ACI versions we set the mode to k8s as the mode is only used by the UI and rancher/k8s are identical in the way they are handled

PR #361, opened 2020-09-07
Automate creation of natgw and routes, plus misc changes for csr integration

jojimt:
- Automatic creation of natgw and route update
 - Clean up of natgw on delete
 - Fetch vrf encap
 - Misc fixes


provision/setup.py
```diff
@@ -28,5 +28,6 @@
           'pyopenssl',
           'ipaddr',
           'MarkupSafe<2.0',
+          'boto3',
```
snaiksat, line 31: Would this need to be packaged in the deb and/or rpm?  If so, we would need to check if and what needs to be added, here:
https://github.com/noironetworks/acc-provision/blob/master/provision/debian/control
and/or here:
https://github.com/noironetworks/acc-provision/blob/master/provision/rpm/acc-provision.spec.in

jojimt: OK, I'll update these

apoorva11029 (approved):

PR #383, opened 2020-10-16
Add flavor for openshift 4.5 on openstack

apoorva11029:
The setting are similar for the 4.4 flavor


snaiksat:
No UT for 4.5 flavor? I know it will be a duplication of the 4.4 UT, but it will be good to add explicitly (if in the future a few things diverge).

snaiksat:
We also need to do the same thing for the OpenShift on ESX flavor.

apoorva11029:
> No UT for 4.5 flavor? I know it will be a duplication of the 4.4 UT, but it will be good to add explicitly (if in the future a few things diverge).

Added.

apoorva11029:
> We also need to do the same thing for the OpenShift on ESX flavor.

There's a UT for that flavor in the code already :)

abhis2112 (approved):

PR #386, opened 2020-10-20
LB IP config for openshift 4.4 ESX flavor

apoorva11029:
User has to add a mandatory config field inside aci_config/vmm_domain/nested_inside section:
installer_provisioned_lb_ip:

This is exported to host-agent configmap field called installer-provisioned-lb-ip


snaiksat:
Should we be updating this:
https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/templates/provision-config.yaml

to add a comment in the nested section.

apoorva11029:
> Should we be updating this:
> https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/templates/provision-config.yaml
> 
> to add a comment in the nested section.

Good point, added!

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -17,6 +17,10 @@ aci_config:
     # type: vmware              # Specify the VMM vendor (supported: vmware)
     # name: myvmware            # Specify the name of the VMM domain
 
+    # Loadbalancer IP used to create duplicate service file for
+    # router-default service in Openshift on ESX case
+    # installer_provisioned_lb_ip: 192.168.18.201
```
snaiksat, line 22: Sorry, one more small nit, can we say this is a "required" field in that case? Also, can we say that this should correspond to the Loadbalancer IP used in the installer?

apoorva11029: Added.

snaiksat (approved):

PR #399, opened 2020-11-05
Adding the disable-multus flag in acc-provision

abhis2112:
--disable-multus flag --> default true for OSP-4.z openstack/ESX releases


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -1044,6 +1044,10 @@ spec:
             - name: NODE_EPG
               value: "{{ config.aci_config.app_profile }}|{{ config.aci_config.nodes_epg }}"
             {% endif %}
+            {% if ( (config.multus.disable) and (config.kube_config.generate_installer_files or config.kube_config.generate_cnet_file) ) %}
```
snaiksat, line 1047: Should this be:

  (not config.multus.disable)

and everywhere below as well.

abhis2112: It's a double negative, the default value is config.multus.disable = True which means the cluster-network-03.yaml file will be added and will disable multus

snaiksat:
Should have mentioned this earlier, but can you update the following file to document the use of this option in the input file:
https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/templates/provision-config.yaml#L82

(it should be in comments like some of the other options)

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -82,3 +82,6 @@ registry:
 
 #drop_log_config:
   # enable: True         # default is False
+
+#disable_multus:
```
snaiksat, line 86: i am not sure if this section name is correct, it should be "multus" right?

abhis2112: so referring to the above lines of code, this is the variable name in which the value is stored from the cmd line flag.

snaiksat: This is an example of the input file, its not related to the CLI. It needs to be what is being read into the config dict, in this case you have used "multus".

abhis2112: right, resolving

apoorva11029 (approved):

PR #403, opened 2020-11-06
Changes for python3 support in debian packaging

apoorva11029:


provision/debian/control
```diff
@@ -2,7 +2,8 @@ Source: acc-provision
 Section: python
 Priority: optional
 Maintainer: Cisco Systems, Inc. <dhami@noironetworks.com>
-Build-Depends: debhelper (>= 9), python-setuptools (>= 3.3), python-all (>= 2.7)
+X-Python-Version: >= 2.6
```
snaiksat, line 5: Let's check if this is the right thing to use here...

snaiksat (approved):

PR #404, opened 2020-11-06
Adding the disable-multus flag in acc-provision

abhis2112:
The flag applied will enable/diable the multus configuration in openshift.
--disable-multus <false> <true>


snaiksat (approved):

snaiksat:
Posting another PR for this.

PR #419, opened 2020-11-26
Move travis builds to bionic (18.04) from xenial (16.04)

tomflynn:


snaiksat:
nit: would be preferable to state "Travis builds" instead of just "builds".

tomflynn:
> nit: would be preferable to state "Travis builds" instead of just "builds".

Done

snaiksat (approved):

PR #420, opened 2020-11-28
Add podif-crd in controller config

tanyatukade:


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -299,6 +299,19 @@ spec:
     served: true
     storage: true
 ---
+apiVersion: apiextensions.k8s.io/v1beta1
```
tomflynn, line 302: Please rework the if checks above to avoid repeating the CRD definition

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -744,6 +757,12 @@ rules:
   - watch
   - get
   - update
+- apiGroups:
+  - "aci.aw"
+  resources:
+  - podifs
+  verbs:
+  - '*'
```
tomflynn, line 765: Same comment here. There's set of podifs permissions for 'use_aci_anywhere_crd' that should should be expanded to all cases

tomflynn (changes requested):

provision/acc_provision/templates/aci-containers.yaml
tanyatukade: removed the check for podif to resuse the crd in non-overlay case as well.

provision/acc_provision/templates/aci-containers.yaml
tanyatukade: done

tanyatukade: in the aci_anywhere check, the permissions are '*'. Should that be expanded?

provision/testdata/base_case_tar/cluster-network-22-ClusterRole-aci-containers:host-agent.yaml
```diff
@@ -86,12 +93,6 @@ rules:
   - list
   - watch
   - get
-- apiGroups:
```
tomflynn, line 89: This change doesn't seem related...merge issue?

tomflynn (commented):
Looks good but please take a look at that one unexpected yaml change

provision/testdata/base_case_tar/cluster-network-22-ClusterRole-aci-containers:host-agent.yaml
tomflynn: Actually this might just be git incorrectly detecting a rename...

snaiksat:
There are some files getting renamed (more specifically the file order number changing) in the cloud tar. I am wondering why? This would typically happen if new resources are getting added which would result in new manifest files those resources.

tanyatukade:
I manually checked for the config.openshift.io config and it is not deleted. It exists in the cluster-network-22 in openshift_45_esx folder.

tanyatukade:
Will open a new PR since resolving conflicts manually for all the testdata files is time consuming.

PR #431, opened 2020-12-09
opflex agent notify socket is not used in k8s deployments. Disable it

tomflynn:


gautvenk (approved):
lgtm

PR #435, opened 2020-12-14
Update to python3 deps

tomflynn:
Point to bionic instead of xenial

### DIFF
--- a/provision/debian/control
+++ b/provision/debian/control
@@ -4,14 +4,14 @@ Priority: optional
 Maintainer: Cisco Systems, Inc. <dhami@noironetworks.com>
 X-Python-Version: >= 3.0
 Build-Depends: debhelper (>= 9), python-setuptools (>= 3.3), python3-all, dh-python
-Standards-Version: 3.9.5
+Standards-Version: 4.1.4
 Homepage: http://github.com/noironetworks/aci-containers/provision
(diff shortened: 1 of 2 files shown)


snaiksat:
have we tried building a deb with these changes?

provision/debian/control
```diff
@@ -4,14 +4,14 @@ Priority: optional
 Maintainer: Cisco Systems, Inc. <dhami@noironetworks.com>
 X-Python-Version: >= 3.0
 Build-Depends: debhelper (>= 9), python-setuptools (>= 3.3), python3-all, dh-python
-Standards-Version: 3.9.5
+Standards-Version: 4.1.4
```
apoorva11029, line 7: Just to confirm, have you reviewed any changes that need to be done for this standards' upgrade?
Based on: https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-standards-version and https://www.debian.org/doc/debian-policy/ch-source.html#s-standardsversion

tomflynn:
> have we tried building a deb with these changes?

Yes

provision/debian/control
tomflynn: No. lintian was complaining so I bumped the version and there were no new lintian complaints. The resulting deb installs fine on Ubuntu 18.04

apoorva11029 (approved):

PR #436, opened 2020-12-14
Corrected if check to create multus files mount

abhis2112:
1. Earlier -> (config.multus.disable), in case config.multus.disable is True, the check should be false
2. Now -> !(config.multus.disable)

(cherry picked from commit fd797afcb1627a4d82957ea9bcc9cfbd598aa3b4)


snaiksat:
The travis gate has passed, but the status was not reported back by travis. Hence, merging it.

snaiksat (approved):

PR #441, opened 2020-12-22
Add flavor for pure overlay setups (no APIC/cAPIC)

tomflynn:
Repurposing the unused, experimental localhost flavor.
This flavor can be used in kind deployments and for basic
overlay deployments.
Initially this flavor does not support automatically enabling
istio or being deployed using the operator.


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -390,10 +390,13 @@ data:
     {
         "flavor": {{config.flavor|json}},
         "log-level": {{config.logging.controller_log_level|json}},
+        {% if config.aci_config.capic or config.kube_config.opflex_mode != "overlay" %}
```
jojimt, line 393: Might be better to use config.flavor != "k8s-overlay".

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -390,10 +390,13 @@ data:
...
         "apic-hosts": {{config.aci_config.apic_hosts|json|indent(width=8)}},
+        {% endif %}
         {% if config.aci_config.capic %}
         "max-csr-tunnels": {{config.net_config.max_csr_tunnels|json|indent(width=8)}},
         {% endif %}
+        {% if config.aci_config.capic or config.kube_config.opflex_mode != "overlay" %}
```
jojimt, line 399: same here

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -405,26 +408,31 @@ data:
...
+        {% endif %}
         {% if config.aci_config.capic %}
         "lb-type": "None",
         {% endif %}
         {% if config.istio_config.install_istio %}
         "install-istio": {{ config.istio_config.install_istio|json }},
         "istio-profile": {{ config.istio_config.install_profile|json }},
         {% endif %}
-        {% if not config.aci_config.capic %}
+        {% if not config.aci_config.capic and config.kube_config.opflex_mode != "overlay" %}
```
jojimt, line 419: same here.

jojimt (commented):
I think it's better to explicitly use config.flavor == "k8s-overlay" in places where we want to avoid adding (c)apic specific config.

provision/acc_provision/templates/aci-containers.yaml
tomflynn: OK I like that suggestion. I'll update the PR shortly

tomflynn: I've updated all checks to use flavor instead of opflex_mode. Let me know if you have any other comments/suggestions

jojimt (approved):

PR #442, opened 2020-12-22
Add contract to allow opflex-agent port to be scraped by prom-server

gautvenk:
- In on-prem setup, port 9612 needs to be provided by node (opflex-agent) and consumed by default epg (in-cluster prom-server)
- When we introduce new filter the order of existing filter can change. Deducing api filter index dynamically (instead of index 7) for dockerucp and rke.

Signed-off-by: Gautam Venkataramanan <gautam.chennai@gmail.com>

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -5070,9 +5070,13 @@ class ApicKubeConfig(object):
                 openshift_flavor_specific_handling(data, items, system_id, old_naming, self.ACI_PREFIX, default_provide_api,
                                                    kube_api_entries, api_filter_prefix, dns_entries, filter_prefix)
             elif flavor == "docker-ucp-3.0":
-                dockerucp_flavor_specific_handling(data, items)
+                dockerucp_flavor_specific_handling(data, items, api_filter_prefix)
             elif flavor == "RKE-1.2.3":
-                rke_flavor_specific_handling(aci_prefix, data, items, self.config["rke_config"])
+                rke_flavor_specific_handling(aci_prefix, data, items, api_filter_prefix, self.config["rke_config"])
+
+        # Adding prometheus opflex-agent contract for all flavors
+        add_prometheus_opflex_agent_contract(data, epg_prefix, contract_prefix, filter_prefix)
+
         self.annotateApicObjects(data, pre_existing_tenant)
         return path, data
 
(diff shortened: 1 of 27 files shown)


shastrinator:
Whether in-cluster prometheus server should be run in the default namespace or not, is not upto us. It is upto the individual installed system. You cannot hold opinions about how it should be installed on managed systems like RKE. Default RKE installation prefers external prometheus server and in-cluster prometheus agent.

gautvenk:
> Whether in-cluster prometheus server should be run in the default namespace or not, is not upto us. It is upto the individual installed system. You cannot hold opinions about how it should be installed on managed systems like RKE. Default RKE installation prefers external prometheus server and in-cluster prometheus agent.

Hi Kiran, the change is to add a contract between kube-default and kube-node for all flavors. Since we expose port 9612 by default, if user adds in-cluster prometheus cluster to pull opflex metrics, then this change avoids the extra contract addition step. For RKE, are u suggesting to not add this contract?

gautvenk:
> Whether in-cluster prometheus server should be run in the default namespace or not, is not upto us. It is upto the individual installed system. You cannot hold opinions about how it should be installed on managed systems like RKE. Default RKE installation prefers external prometheus server and in-cluster prometheus agent.

Good point. Let me discuss with you on this.

gautvenk:
I have added opflex-agent prometheus port 9612 contracts between nodes-system and nodes-default. For now, I have removed these contracts for managed k8s like openshiift, rke, dockerucp.

shastrinator (approved):

provision/acc_provision/apic_provision.py
```diff
@@ -5070,9 +5070,14 @@ def kube_tn(self, flavor):
...
                                                    kube_api_entries, api_filter_prefix, dns_entries, filter_prefix)
             elif flavor == "docker-ucp-3.0":
-                dockerucp_flavor_specific_handling(data, items)
+                dockerucp_flavor_specific_handling(data, items, api_filter_prefix)
             elif flavor == "RKE-1.2.3":
-                rke_flavor_specific_handling(aci_prefix, data, items, self.config["rke_config"])
+                rke_flavor_specific_handling(aci_prefix, data, items, api_filter_prefix, self.config["rke_config"])
+
+        # Not adding prometheus opflex-agent contract for managed k8s environments
+        if vmm_type != "OpenShift" and flavor != "docker-ucp-3.0" and flavor != "RKE-1.2.3":
```
apoorva11029, line 5078: Instead of checking for exact flavor names like docker-ucp-3.0 and RKE-1.2.3, we could have a flag in these flavors which indicates that there's no need for prometheus. This way it will be easier to disable prometheus for future RKE/dockerUCP flavors.

gautvenk: Will do, thanks Apoorva for checking the PR.

gautvenk: Discussed with Sumit and Kiran. Since this is a static port exposed from opflex-agent, we agreed to add contracts between node<-->system and node<-->default for port 9612 for all flavors. I will re-upload PR.

gautvenk:
Tested in fab1 for flavor "kubernetes-1.18" using: acc-provision -f kubernetes-1.18 -o aci-cluster-out-2.yaml -c acc_provision_input.yaml -u admin -p noir0123 --skip-kafka-certs --disable-multus=true -a

APIC had contracts between node<-->system and node<-->default for port 9612

shastrinator (approved):

apoorva11029 (approved):

PR #443, opened 2021-01-04
Add support for eks flavor

jojimt:
Provisioning support for eks and its test code.


provision/acc_provision/flavors.yaml
```diff
@@ -1193,6 +1193,133 @@ flavors:
...
+    config:
+      net_config:
+        max_csr_tunnels: 0
+        csr_list: ""
+        node_subnet: "172.0.0.1/24"
+      cloud:
+        provider: aws
+      kube_config:
+        no_priority_class: True
+        use_openshift_run_level: True
```
snaiksat, line 1209: Do you need to use this?

provision/acc_provision/flavors.yaml
```diff
@@ -1193,6 +1193,133 @@ flavors:
...
+            provided: ["inet-ext"]
+            consumed: ["ul-private", "ul-public"]
+          - name: allow-all-in
+            underlay: True
+            filter: match-all
+            consumed: ["inet-ext"]
+            provided: ["ul-private", "ul-public"]
+          - name: allow-cidr
+            underlay: True
+            filter: match-oshift
```
snaiksat, line 1238: Do you need to use this?

provision/acc_provision/flavors.yaml
```diff
@@ -1193,6 +1193,133 @@ flavors:
...
+                  range: [9000, 9999]
+                - name: ent14
+                  prot: udp
+                  range: [30000, 32767]
+                - name: ent15
+                  prot: icmp
+            
+    status: Experimental
+    hidden: False
+    order: 23
```
snaiksat, line 1322: nit: ideally we would like to have all the experimental flavors listed at the end (but that can be addresses separately)

provision/acc_provision/flavors.yaml
jojimt: Yes, this one I'll keep for now. The requirements are similar.

provision/acc_provision/flavors.yaml
jojimt: Good point. I'll remove it.

provision/acc_provision/flavors.yaml
jojimt: Let me move it.

snaiksat (approved):

PR #455, opened 2021-01-26
Adding the new openshift-4.6-openstack flavor for OSP-4.6 version

abhis2112:


provision/acc_provision/flavors.yaml
```diff
@@ -212,6 +212,132 @@ flavors:
     status: null
     hidden: false
     order: 0
```
snaiksat, line 214: this needs to be updated to 1 (and all the other order numbers need to be incremented by 1).

provision/fix-testdata.sh
```diff
@@ -0,0 +1,329 @@
+rm -rf /tmp/generated*
```
snaiksat, line 1: don't commit this file ;-)

provision/testdata/list_flavors.stdout.txt
```diff
@@ -1,5 +1,6 @@
 INFO: Available configuration flavors:
 INFO: openshift-4.5-openstack:	Red Hat OpenShift Container Platform 4.5 on OpenStack
+INFO: openshift-4.6-openstack:	Red Hat OpenShift Container Platform 4.6 on OpenStack
```
snaiksat, line 3: Once you change the order numbers, this order will also change, and you will just need to swap the above two.

snaiksat (changes requested):

provision/acc_provision/flavors.yaml
abhis2112: noted

provision/fix-testdata.sh
abhis2112: ahh yeah, I was careful with that but seems like ended up adding it in multiple places. Resolving..

apoorva11029 (approved):

snaiksat (approved):

PR #497, opened 2021-04-26
Updating k8s flavors to mount /run/netns

vlella:
Added the flag code config in kube_config of both Kubernetes 1.20 and 1.21 flavors. Modified the flavor argument of get_args() in test_main.py followed by running fix-testdata.sh and test_main.py.

(cherry picked from commit ba021f2d7168fa937164496f37e5996507191282)


snaiksat:
@vlella thanks for posting the backport, the process for fixing the UTs has to be redone for this branch.

snaiksat:
@vlella thanks, this commit has 45 files, but the https://github.com/noironetworks/acc-provision/pull/495/files had 38. Could you check which are the extra files? I just want to be sure we are not accidentally checking in something (doesn't seem like but I would like to double check).

snaiksat:
The commit in the master, does actually show 45 files. Not sure why the PR shows 38 files. Regardless, if I compare this commit and the one in the master, we are touching the same files. So we are good, and I will merge this change.

PR #498, opened 2021-05-06
fixed APIC deployment for preexisting Tenant

pmathy:
If a pre-existing tenant with a name that is not the system_id was specified in the input yml file, the APIC deployment would fail due to the VMM domain association using system_id where tn_name was required.


snaiksat:
Thanks @pmathy for identifying and fixing this issue. Sorry we missed it!

snaiksat (approved):

apoorva11029 (approved):

PR #505, opened 2021-05-15
disable the periodic snat sync

vlella:


provision/acc_provision/templates/provision-config.yaml
```diff
@@ -67,7 +67,8 @@ registry:
 #kube_config:
   # ovs_memory_limit: "20Gi"            # override if needed, default is "1Gi"
   # reboot_opflex_with_ovs: "false"     # override if needed, default is "true"
-
+  # snat_operator:
+        # disable_periodic_snat_global_info_sync: True
```
snaiksat, line 71: This example does not seem to be consistent with the overrides configuration that you have here:
https://github.com/noironetworks/acc-provision/pull/505/files#diff-a4df73be61e65d3f6ba506e9f158411a0e411551e7a3baca74b597a0b63bad94R49

which does not put it in a snat_operator section. Which one is correct, this or that?

provision/testdata/with_overrides.inp.yaml
```diff
@@ -46,6 +46,7 @@ kube_config:
   use_privileged_containers: true
   use_openshift_security_context_constraints: true
   allow_kube_api_default_epg: true
+  disable_periodic_snat_global_info_sync: true
```
snaiksat, line 49: How is it that this does not lead to a change in the follow expected test output:
https://github.com/noironetworks/acc-provision/pull/505/files#diff-a4df73be61e65d3f6ba506e9f158411a0e411551e7a3baca74b597a0b63bad94R49

My suspicion is because you have not put it in the "snat_operator" section (your change in: https://github.com/noironetworks/acc-provision/pull/505/files#diff-25c05340004fec4e2ae7b6d3a37836b486d9522802ef98ecb962d699c0986c25R763 looks into the snat_operator section, which is how it should be).

snaiksat (changes requested):

snaiksat:
Please also update the commit's title and the commit message.

PR #507, opened 2021-05-17
Add CRD to configure droplogs

tanyatukade:


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -680,6 +680,64 @@ spec:
...
+                      type: string
+                    srcMAC:
+                      type: string
+                    destMAC:
+                      type: string
+                    srcPort:
+                      type: integer
+                    destPort:
+                      type: integer
+                    ipProto:
```
shastrinator, line 738: ipProto: is read as an integer from opflex.

tanyatukade: update it

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -1306,6 +1365,17 @@ rules:
...
   - patch
+- apiGroups:
+  - "aci.droplog"
+  resources:
+  - droplogs
+  verbs:
+  - list
+  - watch
+  - get
+  - create
```
tomflynn, line 1376: Does the hostagent really need create/update/delete permissions for droplogs? I know our other CRDs are configured this way now but list/watch/get seems like all that's needed

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -680,6 +680,65 @@ spec:
                     minimum: 1
                     maximum: 1023
 ---
+apiVersion: apiextensions.k8s.io/v1
```
shastrinator, line 683: General comment: There shoudl be only one CR that does enable/disable of droplogs. If this is not possible with CRD and web-hook, configmap may be a better option

provision/acc_provision/templates/aci-containers.yaml
tanyatukade: Yes, list/watch/get should be enough. Will update that, thanks!

provision/acc_provision/templates/aci-containers.yaml
tanyatukade: As discussed, we will have a CRD with node selector and selectively enable droplog. We keep track of all CRs defined and then merge. If any CR has the feature disabled, then it’s disabled irrespective of other CRs.

tanyatukade:
As discussed, we will have two separate CRDs. One CRD is used to enable/disable drop logging feature using node selectors. And the other CRD will have filters. I will update this PR of kind "DropLogEnable" and remove the filters as this will only used to enable/disable. Will create a new PR shortly for the 2nd CRD of kind "DropLogPruning".

tanyatukade:
Updated the PR with two CRDs

shastrinator (dismissed):

tanyatukade:
Resolving conflicts

tomflynn (approved):

PR #508, opened 2021-05-17
 Adding config option to disable periodic SNAT sync in acc-provision …

vlella:
…input

The disable_periodic_snat_global_info_sync field in the nested snat_operator field in kube_config is optional that is the user is not required to set it via acc-provision input field. It's by default false unless user sets it to True.

(cherry picked from commit 1056fd0c1818a66d73ddee7f63866a317a5931b0)


snaiksat:
I think you may not have regenerated the expected test files for this branch. Merely back-porting is not going to work since the image tags are different in this branch (hence the tests are failing).

PR #513, opened 2021-05-25
To make wait_for_network duration configurable

anmol372:


provision/acc_provision/acc_provision.py
```diff
@@ -172,6 +172,7 @@ def config_default():
             "interface_mtu": None,
             "second_kubeapi_portgroup": False,
             "disable_wait_for_network": False,
+            "wait_for_network_duration": 210,
```
snaiksat, line 175: Per earlier suggestion, can we change this to "duration_wait_for_network" (although I know the current name reads better)? I would like to keep the consistency with the naming convention of the related configuration "disable_wait_for_network". Else, we could change "disable_wait_for_network" to "wait_for_network_disable" (but then it will have to be changed in a number of places).

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -59,6 +59,11 @@ net_config:
                                    # CNI should not wait for datapath to be
                                    # ready before returning success for pod
                                    # interface creation
+ #duration_wait_for_network: 210  # Duration in seconds that ACI should wait
```
snaiksat, line 62: this example is not consistent with the value you are reading, suggestion is to change that, but either way it should be consistent

provision/acc_provision/acc_provision.py
```diff
@@ -172,6 +172,7 @@ def config_default():
             "interface_mtu": None,
             "second_kubeapi_portgroup": False,
             "disable_wait_for_network": False,
+            "duration_wait_for_network": "210",
```
snaiksat, line 175: I don't think needs to be a string? (It's also not consistent with your example later in the patch).

provision/acc_provision/acc_provision.py
```diff
@@ -415,7 +417,8 @@ def config_adjust(args, config, prov_apic, no_random):
             "node_network_gateway": cidr_split(node_subnet)[5],
             "pod_network": normalize_cidr(pod_subnet),
             "node_network": normalize_cidr(node_subnet),
-            "disble_wait_for_network": disable_wait_for_network,
+            "disable_wait_for_network": disable_wait_for_network,
+            "duration_wait_for_network": str(duration_wait_for_network),
```
snaiksat, line 421: I think this does not need to be explicitly cast into a string

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -1507,6 +1507,9 @@ spec:
             {% if config.net_config.disable_wait_for_network %}
             - name: DISABLE_WAIT_FOR_NETWORK
               value: 'True'
+            {% else %}
+            - name: DURATION_WAIT_FOR_NETWORK
+              value: {{'\"' + config.net_config.duration_wait_for_network + '\"'}}
```
snaiksat, line 1512: Why do we need to put this in \"\"? The DURATION_WAIT_FOR_NETWORK env var will take the integer value of config.net_config.duration_wait_for_network and which will automatically be available to the launch-hostagent script as a string (and which we are taking care of converting to an integer in the script).

snaiksat (changes requested):
We do need to update the overrides test case, and which will require update to the testdata files.

provision/acc_provision/templates/provision-config.yaml
anmol372, line 62: The duration_wait_for_network value must be a string, since that is what is expected in aci_deployment.yaml.

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -59,6 +59,11 @@ net_config:
                                    # CNI should not wait for datapath to be
                                    # ready before returning success for pod
                                    # interface creation
+ #duration_wait_for_network: "210"  # Duration in seconds that ACI should wait
```
snaiksat, line 62: Since we changed https://github.com/noironetworks/acc-provision/pull/513/files#diff-d18edd87a3bcaa21be667d28554472484dbc49b6c8839bfddec4c68545f9c559R175 , the above example should not be in quotes, right?

snaiksat: Also, if you update the overrides the UT, these things will get tested automatically.

provision/acc_provision/templates/aci-containers.yaml
snaiksat: Change config.net_config.duration_wait_for_network to: config.net_config.duration_wait_for_network|json

snaiksat (approved):

provision/testdata/.acirc
```diff
@@ -1,2 +1,2 @@
-|public subnets|: [u'subnet-06cdb0b67bf86fcbf', u'subnet-0814cdfa1911664e7']
-|private subnets|: [u'subnet-093bf9479f50168f7', u'subnet-00a54d51ba268fe29']
+|public subnets|: ['subnet-06cdb0b67bf86fcbf', 'subnet-0814cdfa1911664e7']
```
snaiksat, line 1: this file should not be committed

snaiksat:
UTs are passing, ignoring the coveralls configuration error because of which Travis CI is failing.

snaiksat (approved):

PR #514, opened 2021-05-27
Add EPRegistry param back in place for podif CRD

tanyatukade:
- The "ep_registry" param was removed to enable ERSPAN feature in on-prem case. This caused issues with the way podifs are processed at scale. This commit reverts the change and puts the param back in place.

- To enable ERSPAN feature, we need to derive the config from acc-provision input file. An example of the config is below:
```
kube_config:
  …
  ep_registry: k8s
```


tomflynn (approved):

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -71,6 +71,7 @@ registry:
 #kube_config:
   # ovs_memory_limit: "20Gi"            # override if needed, default is "1Gi"
   # reboot_opflex_with_ovs: "false"     # override if needed, default is "true"
+  # ep_registry: "k8s"                  # default is null, set to "k8s" to enable ERSPAN feature
```
snaiksat, line 74: the quotes are not required here (arguably the earlier examples are also not accurate)

PR #523, opened 2021-06-11
Flag added as Running UTs modifies .acirc file

vlella:


provision/acc_provision/cloud_provision.py
```diff
@@ -169,12 +169,13 @@ def Run(self, flavor_opts, kube_yaml_gen_func):
         output_tar = self.args.output_tar
         operator_cr_output_file = self.args.aci_operator_cr
         gen(self.config, output_file, output_tar, operator_cr_output_file)
-        if self.args.flavor == "aks":
-            self.print_aks_setup()
-        elif self.args.flavor == "eks":
-            self.print_eks_setup()
-        else:
-            self.print_openshift_setup()
+        if not self.args.test_run:
```
snaiksat, line 172: You will need to first check if test_run is present in args.

vlella: Cool, makes sense I guess -> if not (self.args.test_run in self.args): OR if self.args.test_run not in self.args:

provision/acc_provision/cloud_provision.py
```diff
@@ -169,12 +169,13 @@ def Run(self, flavor_opts, kube_yaml_gen_func):
         output_tar = self.args.output_tar
         operator_cr_output_file = self.args.aci_operator_cr
         gen(self.config, output_file, output_tar, operator_cr_output_file)
-        if self.args.flavor == "aks":
-            self.print_aks_setup()
-        elif self.args.flavor == "eks":
-            self.print_eks_setup()
-        else:
-            self.print_openshift_setup()
+        if not (self.args.test_run in self.args):
```
snaiksat, line 172: if "test_run" not in self.args:

snaiksat (changes requested):
pl see inline

snaiksat (dismissed):

snaiksat:
DNM, @vlella is still checking

snaiksat:
Is there an update on this?

vlella (commented):
The logic looks good @snaiksat as discussed before in discussion but the acirc file still keeps on coming when we do git status. I rechecked it again now the problem persists and trying to figure out what's missing.

provision/acc_provision/cloud_provision.py
```diff
@@ -169,12 +169,14 @@ def Run(self, flavor_opts, kube_yaml_gen_func):
...
         operator_cr_output_file = self.args.aci_operator_cr
         gen(self.config, output_file, output_tar, operator_cr_output_file)
-        if self.args.flavor == "aks":
-            self.print_aks_setup()
-        elif self.args.flavor == "eks":
-            self.print_eks_setup()
-        else:
-            self.print_openshift_setup()
+        if "test_run" in self.args:
+            if not self.args.test_run:
```
snaiksat, line 173: if "test_run" in self.args and not self.args.test_run:

provision/acc_provision/cloud_provision.py
```diff
@@ -169,12 +169,13 @@ def Run(self, flavor_opts, kube_yaml_gen_func):
         output_tar = self.args.output_tar
         operator_cr_output_file = self.args.aci_operator_cr
         gen(self.config, output_file, output_tar, operator_cr_output_file)
-        if self.args.flavor == "aks":
-            self.print_aks_setup()
-        elif self.args.flavor == "eks":
-            self.print_eks_setup()
-        else:
-            self.print_openshift_setup()
+        if "test_run" in self.args and not self.args.test_run:
```
snaiksat, line 172: this still does not work, because now when you actually run acc-provision the files are never going to get generated!

I still dont understand why the earlier suggestion did not work:

    if not "test_run" in self.args:

vlella: hmm okay, 
the earlier suggestion as suggested by you which can be written in 3 ways in python syntax which all mean same -
if "test_run" not in self.args: , if not ("test_run" in self.args): , if not "test_run" in self.args:
for all of these 3 I tried, the acirc file get's generated but I couldn't logically understand why it's getting generated though.

provision/acc_provision/test_main.py
```diff
@@ -791,6 +791,7 @@ def get_args(**overrides):
         # configuration which would otherwise be discovered from
         # the APIC
         "infra_vlan": None,
+        "test_run": False,
```
snaiksat, line 794: I would imagine test_run should be true for all the tests, so I would have expected it to be set to True here (and would not need to be set to True later).

provision/acc_provision/cloud_provision.py
snaiksat: this comment is still not addressed...

snaiksat (changes requested):

provision/acc_provision/cloud_provision.py
```diff
@@ -169,12 +169,13 @@ def Run(self, flavor_opts, kube_yaml_gen_func):
         output_tar = self.args.output_tar
         operator_cr_output_file = self.args.aci_operator_cr
         gen(self.config, output_file, output_tar, operator_cr_output_file)
-        if self.args.flavor == "aks":
-            self.print_aks_setup()
-        elif self.args.flavor == "eks":
-            self.print_eks_setup()
-        else:
-            self.print_openshift_setup()
+        if "test_run" not in self.args._asdict():
```
snaiksat, line 172: Per my earlier request, was this tested with the standalone acc-provision tool (when its not being invoked from the tests)? I think this is still going to create a problem when invoked from the acc-provision tool. While "args" is a namedtuple in the tests, I don't think its a namedtuple when the acc-provision tool runs (and if not, it won't have the _asdict() method). You will need to track the "args" definition in the main().

provision/acc_provision/cloud_provision.py
```diff
@@ -169,12 +169,22 @@ def Run(self, flavor_opts, kube_yaml_gen_func):
...
+                else:
+                    self.print_openshift_setup()
+        except AttributeError:
+            if "test_run" not in vars(self.args):
+                if self.args.flavor == "aks":
+                    self.print_aks_setup()
+                elif self.args.flavor == "eks":
+                    self.print_eks_setup()
+                else:
+                    self.print_openshift_setup()
```
snaiksat, line 187: There is unnecessary code repetition here. "test_run" is only ever going to be part of args when we are running tests, and in that case the prints should not be called, otherwise the prints should be called.

vlella: Makes sense, thanks for the suggestion. I have fixed it and tested it on my VM it works smoothly now.

PR #527, opened 2021-06-21
CRDs for network policy (including dns aware policy)

anmol372:
1. Added CRD to support creation of networkpolicies.aci.netpol CustomResource
2. Added CRD to support creation of dnsnetworkpolicies.aci.dnsnetpol CustomResource
3. Added RBAC permissions to controller and host-agent ClusterRole to allow managemnet of the new custom resources.


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -492,6 +492,442 @@ spec:
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.5.0
```
shastrinator, line 497: Dont need this annotation. Same comment for further CRDs

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -492,6 +492,442 @@ spec:
...
+                        matchNames:
+                          items:
+                            type: string
+                          type: array
+                      required:
+                      - matchNames
+                      type: object
+                  required:
+                  - enableLogging
+                  - toFqDn
```
shastrinator, line 686: just more clear if toDnsName were used, since this can have wildcards. But this is a comment on the code.

anmol372: This would need to be changed in aci-containers patch as well. Should we do it?

shastrinator: Its ok, we can continue with this.

provision/acc_provision/templates/aci-containers.yaml
anmol372: Done

shastrinator (approved):

snaiksat:
@anmol372 thanks, can we please add a little more information in the commit message as to what we are doing and why we are doing it? Since CRDs are being defined here, we need to have more details. Also in the commit title message, we need to say "enhanced network policy" or something similar.

apoorva11029 (approved):

PR #530, opened 2021-06-25
Allow specifying multiple subnets for pod networking

vlella:


provision/acc_provision/acc_provision.py
```diff
@@ -414,10 +418,9 @@ def config_adjust(args, config, prov_apic, no_random):
         },
         "net_config": {
             "infra_vlan": infra_vlan,
-            "gbp_pod_subnet": "%s/%s" % (cidr_split(pod_subnet)[2], cidr_split(pod_subnet)[4]),
+            "gbp_pod_subnet": "%s/%s" % (cidr_split(pod_subnets[0])[2], cidr_split(pod_subnets[0])[4]),
```
snaiksat, line 421: Please add a TODO comment above this line stating that for the overlay mode we are currently not supporting multiple subnets, hence you are assuming that there is only one subnet. Supporting multiple subnets on gbp-server side will require more changes.

provision/acc_provision/acc_provision.py
```diff
@@ -494,6 +479,43 @@ def config_adjust(args, config, prov_apic, no_random):
...
+                            "gateway": cidr_split(pod_subnet)[2],
+                            "routes": [
+                                {
+                                    "dst": config_set_dst(pod_subnet),
+                                    "gw": cidr_split(pod_subnet)[2],
+                                }
+                             ]
+                          }
+                        ]
+                )
```
snaiksat, line 508: This will result in multiple pod_ip_pool and pod_network key:value items in the kube_config_object. Instead we need to add each subnet as a new entry in each of the lists corresponding to the values of the keys "pod_ip_pool" and "pod_network", e.g.:

     "pod_ip_pool": [ {}, {}, ..]

Similarly, in the net_config_object below.

Also, there is a "," missing between the items "pod_ip_pool" and "pod_network".

provision/testdata/base_case.apic.txt
```diff
@@ -775,17 +775,17 @@ None
...
                                 "attributes": {
-                                    "ip": "10.2.0.1/16",
+                                    "tnFvCtxName": "kube",
                                     "annotation": "orchestrator:aci-containers-controller"
                                 }
                             }
                         },
                         {
-                            "fvRsCtx": {
+                            "fvRsBDToOut": {
```
snaiksat, line 786: This doesn't seem right because now we have completely lost the fvSubnet. Similarly in test cases below too.

vlella: Just checked of the total 30 apic.txt files in testdata the below mentioned 20 names are the ones without fvSubnet - 

base_casea.apic.txt
base_casea_ipv6.apic.txt
flavor_dockerucp.apic.txt
flavor_openshift_310.apic.txt
flavor_openshift_311.apic.txt
nested-elag.apic.txt
nested-portgroup.apic.txt
nested-vlan.apic.txt
nested-vxlan.apic.txt
vlan_case.apix.txt
with_comments.apic.txt
with_interface_mtu.apic.txt
with_new_naming_convention.apic.txt
with_new_naming_convention_dockerucp.apic.txt
with_new_naming_convention_openshift.apic.txt
with_overrides.apic.txt
with_preexisting_tenant.apic.txt
with_tenant_l3out.apic.txt
with_wait_for_network.apic.txt

so in children of fvBD with kube-pod-bd as attribute I guess I need to add this chunk of code after comma in all these remaining 20 files of apic.txt - 
,
{
    "fvSubnet": {
        "attributes": {
            "ip": "10.2.0.1/16",
            "scope": "public",
            "annotation": "orchestrator:aci-containers-controller"
        }
    }
}

LGTM now I guess but does it sounds good to you, Sumit?

provision/testdata/cloud_tar/cluster-network-03-config.yaml
```diff
@@ -6,7 +6,9 @@ metadata:
 spec:
   disableMultiNetwork: true
   clusterNetwork:
-  - cidr: "14.3.0.1/16"
+  - cidr: [
+    "14.3.0.1/16"
+]
```
snaiksat, line 11: This might be a problem - I don't think OpenShift cluster network operator supports a list:
https://github.com/openshift/cluster-network-operator

I think we will not be able to support multiple subnets for the OpenShift case, meaning, we throw an error (at least for now) if multiple subnets are configured in the case of the OpenShift flavor. Secondly, the above would always be just a single subnet, not a list.

@abhis2112 any thoughts on the above?

abhis2112: Correct Sumit. The way the network operator supports multiple  pod networks is like below:

ClusterNetwork: []configv1.ClusterNetworkEntry{
		{
			CIDR:       "10.0.0.0/22",
			HostPrefix: 24,
		},
		{
			CIDR:       "10.2.0.0/22",
			HostPrefix: 23,
		},
	},


The CIDR can't be a list. It's a single subnet string. Any right now we are supporting a single subnet and also in the case of multus is on a different note and not configured like aforementioned.

snaiksat:
This is has been subsumed by another PR.

PR #532, opened 2021-06-29
Adding openshift-4.7-openstack flavor for OCP-4.7 in acc-provision

abhis2112:


provision/acc_provision/flavors.yaml
```diff
@@ -123,6 +123,132 @@ flavors:
...
+        vmm_domain:
+          type: OpenShift
+      istio_config:
+        install_istio: False
+      multus:
+        disable: True
+    options: *anchor1
+    status: null
+    hidden: false
+    order: 0
```
apoorva11029, line 251: Please increment the order numbers for the rest of the flavors accordingly

apoorva11029 (changes requested):

provision/acc_provision/flavors.yaml
abhis2112: yes, sure nice catch . thank you, I will resolve right away

snaiksat (approved):

apoorva11029 (approved):

PR #533, opened 2021-07-02
sriov daemonset and config map as a part of acc_provision

Bhavanaashok33:
Now that we are planning to support ovs hardware offloading using smart nics, sriov device plugin will be deployed as a  daemonset for discovering and advertising SR-IOV virtual functions which are available on k8s clusters. Provided we enable the "sriov_config" knob in the acc_provision input file(overlay and on-prem), sriov network plugin daemonset and config map will be generated as a part of aci_deployment file.  

acc_provision changes are tested on the setup and verified the generated yaml file from test cases thats been added.

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -234,6 +234,9 @@ def config_default():
         },
         "multus": {
             "disable": True,
+        },
+        "sriov_config": {
+            "enable": False,
         }
     }
     return default_config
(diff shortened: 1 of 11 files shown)


Bhavanaashok33:
This PR has no additional changes compared to the previous PR. It is just a new branch with similar changes. @snaiksat  Please take a look when possible

snaiksat (approved):

PR #535, opened 2021-07-09
Introduce NodePodIF CRD

tanyatukade:
- The original PodIF CRD is left intact and will only be used in the pure overlay flavor.
- Refer: https://github.com/noironetworks/support/issues/1539
- The new NodePodIF CRD has a list of podifs for all pods on each node. Hence each node will have only 1 NodePodIF CR. Therefore the API requests happen per node rather than per pod.
- Moved PodIF attributes from Status to Spec field.
- The consumers of podif such as the ERSPAN feature will use nodepodifs instead of podifs


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -167,51 +167,54 @@ spec:
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  name: podifs.aci.aw
+  name: nodepodifs.aci.aw
```
snaiksat, line 170: Since our interim plan is to keep podifs intact such that gbpserver keeps working, we should be adding nodepodifs, but not removing podifs.

tanyatukade: Updated the PR

snaiksat:
nit: can we change the commit title to say something to the effect of "Introducing nodepodifs CRD" since podif CRD is no longer being reworked in this PR. The commit message can also be updated to reflect that change (that we are moving the consumers of podif such as erspan feature to use nodepodifs instead of podifs, but podifs itself is being left intact).

tanyatukade:
> nit: can we change the commit title to say something to the effect of "Introducing nodepodifs CRD" since podif CRD is no longer being reworked in this PR. The commit message can also be updated to reflect that change (that we are moving the consumers of podif such as erspan feature to use nodepodifs instead of podifs, but podifs itself is being left intact).

Updated the commit message along with the title

snaiksat (approved):

PR #540, opened 2021-07-21
Add config knob to enable nodepodif

tanyatukade:
- By default the knob is set to false and has to be provisioned via acc_provision config file.
- Features such as ERSPAN will rely on this.

- To enable this set the below config in acc_provision_input.yaml 

```
nodepodif_config:
  enable: True
```
- Corresponding aci-containers change: https://github.com/noironetworks/aci-containers/pull/833


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -1780,12 +1787,14 @@ rules:
   - watch
   - get
   - update
+{% if config.nodepodif_config.enable %}
```
snaiksat, line 1790: My suggestion is that we dont change any of this, we just take the flag from the user and relay it to the host-agent config. That way, the CRD and the necessary RBAC is already in place if this flag is flipped let's say directly in the config-map.

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -1967,12 +1976,14 @@ rules:
   - get
   - list
   - watch
+{% if config.nodepodif_config.enable %}
```
snaiksat, line 1979: same here

provision/acc_provision/templates/aci-containers.yaml
tanyatukade: sure, that makes sense, will update the PR shortly. Thanks for the review.

snaiksat (approved):

apoorva11029 (changes requested):
Can we move the new UT to override testcase?

tanyatukade:
> Can we move the new UT to override testcase?

Moved it to override testcase.

apoorva11029 (approved):

PR #541, opened 2021-07-21
Introduce an operator to manage CNI lifecycle

apoorva11029:
Added acc-provision-operator container inside the acicnioperator pod to manage upgrades for the CNI using the accprovisioninput CRD

The CRD currently supports modification of input fields related to kubernetes resources like pod image tags, ovs pod memory lists, image pull policy etc - full list can be found under provision/acc_provision/templates/acc-provision-crd.yaml

This operator can also upgrade the CNI version if the user deploys the updated version of the operator image


tanyatukade (approved):
LGTM

provision/acc_provision/versions.yaml
```diff
@@ -45,6 +45,7 @@ versions:
     opflex_server_version: 5.1.1.0.1ae238a
     gbp_version: 5.1.1.0.1ae238a
   5.2:
+    acc_provision_operator_version: newacioperator
```
snaiksat, line 48: this has probably crept in from the dev env, it should be the same as others below...

provision/acc_provision/test_main.py
```diff
@@ -786,6 +798,7 @@ def get_args(**overrides):
         "skip_kafka_certs": True,
         "upgrade": False,
         "disable_multus": 'true',
+        "operator_mode": False,
```
snaiksat, line 801: I think we should put a comment here that at some point we need to make the default as True so that all new test cases are automatically run with this. In fact, it would be good to evaluate how much extra effort is needed to make this default now.

snaiksat (changes requested):

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    subresources:
+      status: {}
+    schema:
+      openAPIV3Schema:
+        description: accprovisioninput owns the lifecycle of ACI objects in the cluster
```
snaiksat, line 22: This should be updated to the relevant description (its currently copy-pasted from aci-operators-crd).

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+        properties:
+          apiVersion:
+            type: string
+          kind:
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: AccProvisionInputSpec defines the desired spec for ACI Objects
```
snaiksat, line 31: Ditto

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+                      aci_containers_host_version:
+                        type: string
+                      cnideploy_version:
+                        type: string
+                      opflex_agent_version:
+                        type: string
+                      openvswitch_version:
+                        type: string
+                      gbp_version:
+                        type: string
```
snaiksat, line 81: aci_containers_operator_version seems to be missing.

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+                  multus:
+                    type: object
+                    properties:
+                      disable:
+                        type: boolean
+                  drop_log_config:
+                    type: object
+                    properties:
+                      enable:
+                        type: boolean
```
snaiksat, line 107: some new configurations have been recently added:
sriov_config
nodepodif_config

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+                      sync_login:
+                        type: object
+                        properties:
+                          certfile:
+                            type: string
+                          keyfile:
+                            type: string
+                      client_ssl:
+                        type: boolean
+                  net_config:
```
snaiksat, line 50: When comparing with the following, some fields are missing:
https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/acc_provision.py#L161-L176
(namely, service_monitor_interval, disable_wait_for_network, duration_wait_for_network)

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+                    properties:
+                      ovs_memory_limit:
+                        type: string
+                      use_external_service_ip_allocator:
+                        type: boolean
+                      use_privileged_containers:
+                        type: boolean
+                      use_openshift_security_context_constraints:
+                        type: boolean
+                      allow_kube_api_default_epg:
```
snaiksat, line 119: seem like this would be an immutable field, so shouldn't be here?

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+                  kube_config:
+                    type: object
+                    properties:
+                      ovs_memory_limit:
+                        type: string
+                      use_external_service_ip_allocator:
+                        type: boolean
+                      use_privileged_containers:
+                        type: boolean
+                      use_openshift_security_context_constraints:
```
snaiksat, line 117: seem like this would be an immutable field, so shouldn't be here?

provision/acc_provision/templates/acc-provision-crd.yaml
```diff
@@ -0,0 +1,150 @@
...
+                    type: object
+                    properties:
+                      enable:
+                        type: boolean
+                  kube_config:
+                    type: object
+                    properties:
+                      ovs_memory_limit:
+                        type: string
+                      use_external_service_ip_allocator:
```
snaiksat, line 113: seem like this would be an immutable field, so shouldn't be here?

snaiksat (changes requested):

provision/acc_provision/acc_provision.py
```diff
@@ -1280,6 +1286,9 @@ def parse_args(show_help):
     parser.add_argument(
         '--disable-multus', default='true', metavar='disable_multus',
         help='true/false to disable/enable multus in cluster')
+    parser.add_argument(
+        '--operator-mode', default=False,
```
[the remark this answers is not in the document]
apoorva11029: Since this operator-mode flag is to be used only by the ansible operator internally, maybe it makes more sense to keep it False by default for our testcases?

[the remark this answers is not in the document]
snaiksat: I am not suggesting that it be changed. I am suggesting a comment be added in the code to clarify what is the purpose of this, and I have also suggested what that comment should be (so that its clear what my suggestion is). I had not put the comment in quotes earlier, now I have (in case that was the source of confusion).

provision/acc_provision/test_main.py
apoorva11029: Since this operator-mode flag is to be used only by the ansible operator internally, maybe it makes more sense to keep it False by default for our testcases?

I meant to post this reply here

provision/acc_provision/acc_provision.py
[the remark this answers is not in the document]
apoorva11029: I meant to post this reply to your comment before this, corrected it now

Yes I agree we need the description.

provision/acc_provision/templates/acc-provision-crd.yaml
apoorva11029: Do you mean use_external_service_ip_allocator? That's used in controller config: https://github.com/noironetworks/acc-provision/blob/0754b53ffdccfae3a7edb16e54216ba5a5f2896a/provision/acc_provision/templates/aci-containers.yaml#L1407

provision/acc_provision/templates/acc-provision-crd.yaml
apoorva11029: Do you meant use_openshift_security_context_constraints?
It's used to generate kube yaml as well https://github.com/noironetworks/acc-provision/blob/0754b53ffdccfae3a7edb16e54216ba5a5f2896a/provision/acc_provision/templates/aci-containers.yaml#L2013

provision/acc_provision/templates/acc-provision-crd.yaml
apoorva11029: Right, moved

provision/acc_provision/templates/acc-provision-crd.yaml
snaiksat: Doesn't this CRD represent just the mutable fields? if so, I don't think we should allow changing the value of use_external_service_ip_allocator.

provision/acc_provision/templates/acc-provision-crd.yaml
snaiksat: Ditto

provision/acc_provision/test_main.py
snaiksat: Okay for now, lets make a note of this point, and discuss it as a follow up to this patch.

provision/acc_provision/templates/acc-provision-crd.yaml
apoorva11029: Got it,removed

provision/acc_provision/templates/acc-provision-crd.yaml
apoorva11029: Removed

snaiksat (approved):

snaiksat:
Merging based on earlier LTGM comments from other reviewers.

snaiksat (approved):

PR #545, opened 2021-08-02
Introducing flavor OCP 4.6 on baremetal

apoorva11029:
The new flavor is similar to the OpenShift on OpenStack 4.6 flavor as far as the contracts are concerned.
It differs from that flavor in the programming of node EPG- . In this case we will have the user provide the kube-api-vlan, and use that to program the infraRsFuncToEpg in access(untagged) mode
By default, set the mode to regular


provision/acc_provision/flavors.yaml
```diff
@@ -375,7 +375,134 @@ flavors:
...
+            etherT: ip
+            prot: udp
+            stateful: "no"
+        vmm_domain:
+          type: OpenShift
+      istio_config:
+        install_istio: False
+      multus:
+        disable: True
+    options: *anchor1
```
snaiksat, line 502: This should be removed since we do want the kube_api_vlan to be configured.

provision/testdata/flavor_openshift_46_baremetal.apic.txt
```diff
@@ -0,0 +1,1867 @@
+/api/mo/uni/infra/vlanns-[kube-pool]-static.json
```
snaiksat, line 1: Currently the infraRsFuncToEpg is not being configured at all for node EPG. Please revisit and validate this file again after the "anchor1" option is removed from this flavor's options per suggestion above.

snaiksat (changes requested):
In commit message:
register -> regular

provision/testdata/flavor_openshift_46_baremetal.apic.txt
vlella: just noticed that acirc is being checked in, I hope it's by mistake.

apoorva11029:
Thanks @snaiksat and @vlella, I have made the requested changes- pls take a look

provision/testdata/flavor_openshift_46_baremetal.inp.yaml
```diff
@@ -0,0 +1,50 @@
...
+    keyfile: user.key
+  vmm_domain:
+    encap_type: vxlan
+    mcast_range:
+        start: 225.2.1.1
+        end: 225.2.255.255
+  custom_epgs:
+    - group1
+    - group2
+  preexisting_kube_bd: neutron-bd-test
```
snaiksat, line 28: this should not be used here...

apoorva11029: good catch, fixing

snaiksat (approved):

PR #546, opened 2021-08-04
Introduce leaf opflex cert and ssl options for new APIC version

apoorva11029:
For APIC >= 5.2(3), set leafOpflexpAuthenticateClients and leafOpflexpUseSsl instead of opflexpAuthenticateClients and opflexpUseSsl

Also make the apic_version field used in provision code a string, for exmple 5.2(3e) APIC version would be coverted to 5.2.3 internally


provision/acc_provision/acc_provision.py
```diff
@@ -123,7 +123,7 @@ def config_default():
     # Default values for configuration
     default_config = {
         "aci_config": {
-            "apic_version": 1.0,
+            "apic_version": "1.0",
```
snaiksat, line 126: Instead of setting the apic_version as "a.b.c", can we read it in the actual format "a.b(cz)" (e.g. "5.2(3e)") which you expect to get from the APIC?

snaiksat (changes requested):

provision/acc_provision/acc_provision.py
apoorva11029: This field aci_config.apic_version will always look like a.b.c since we process the incoming version from APIC

provision/acc_provision/apic_provision.py
```diff
@@ -2345,6 +2362,37 @@ def opflex_cert(self):
         self.annotateApicObjects(data)
         return path, data
 
+    def opflex_cert_leaf(self):
```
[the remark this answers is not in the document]
apoorva11029: Sure, adding

provision/acc_provision/apic_provision.py
```diff
@@ -305,13 +306,22 @@ def unprovision(self, data, system_id, tenant, vrf_tenant, cluster_tenant, old_n
...
+
     def get_apic_version(self):
         path = "/api/node/class/firmwareCtrlrRunning.json"
-        version = 1.0
+        version = "1.0"
         try:
             data = self.get_path(path)
             versionStr = data['firmwareCtrlrRunning']['attributes']['version']
-            version = float(versionStr.split('(')[0])
+            version = self.process_apic_version_string(versionStr)
```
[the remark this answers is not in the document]
apoorva11029: Yes

provision/acc_provision/acc_provision.py
snaiksat: My understanding is that we are using apic_version primarily for UT purposes. So why can we not interpret it in the format that we expect from APIC? That helps us to test the real APIC format even in UTs, no?

apoorva11029: Based off our offline discussion, created an issue for this.

snaiksat (approved):

PR #551, opened 2021-08-17
Remove ep_registry congig knob for on-prem case

tanyatukade:
The new CRD "NodePodIF" is introduced to handle the CRUD of podifs in one-prem case.
ep_registry knob is only used in overlay case today


tanyatukade:
DNM: Needs testing

snaiksat (approved):

PR #560, opened 2021-09-11
Update provision-config.yaml with operator_managed_config field

abhis2112:


provision/acc_provision/templates/provision-config.yaml
```diff
@@ -1,6 +1,8 @@
 #
 # Configuration for ACI Fabric
 #
+operator_managed_config: 
+  enable_updates: false
```
snaiksat, line 5: This file is an example to the user as to how to change the default configuration, so we don't provide the default configuration (because there is already in the system). The above should be:

      #
      # Configuration for ACI CNI Operator
      #

      #
      # operator_managed_config:
      #   enable_updates: True    # Default is False

Also, this is not "ACI Config", so please move this to L86 (after the kube_config and before istio_config sections).

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -81,6 +81,10 @@ registry:
   # reboot_opflex_with_ovs: "false"     # override if needed, default is "true"
   # snat_operator:
         # disable_periodic_snat_global_info_sync: True
+
```
snaiksat, line 84: nit: in my earlier commit I suggested adding a comment which explains what this section does. That is not present here:

 #
  # Configuration for ACI CNI Operator
  #

abhis2112: i couldn't found any suggestion there :

This file is an example to the user as to how to change the default configuration, so we don't provide the default configuration (because there is already in the system). The above should be:

  #
  # Configuration for ACI CNI Operator
  #

  #
  # operator_managed_config:
  #   enable_updates: True    # Default is False


Also, this is not "ACI Config", so please move this to L86 (after the kube_config and before istio_config sections).

snaiksat (approved):

PR #561, opened 2021-09-11
To add operator_managed_config section to provision-config

vlella:


snaiksat:
This was already addressed here: https://github.com/noironetworks/acc-provision/pull/560

PR #564, opened 2021-09-15
Bumping release to 6.0.0.0

anmol372:


snaiksat:
Thanks @anmol372 . We would have regenerate the test files though. :-(

snaiksat:
@vlella please take note of this PR for feature, on the release branches (eg kwai, kmr2, etc) also we do something similar. We did one yesterday in fact on kmr2.

snaiksat:
@anmol372 the tests are passing because 6.0 version is not the default. We would have to change in in the flavors.yaml (then the tests would fail and we would need to update the test data).

anmol372:
> Thanks @anmol372 . We would have regenerate the test files though. :-(

I ran fix-tesdata but no new files generated, I think I need to update the tests to use the release as well. Will update soon.

anmol372:
Build is failing as coveralls.io api is down. Monitoring it and will restart build once that's back up.

anmol372:
coveralls service is back and CI is passing.

snaiksat (approved):

snaiksat (changes requested):
nit (but important) - can you change the commit message to:
Bumping release to 6.0.0.0

snaiksat (approved):

PR #578, opened 2021-11-18
Add cko new flavor and calico_config for the cko PoC

tanyatukade:
- Intial draft for calico_config in acc_provision input file
- I need to check if service_cluster_ips and service_external_ips overlap with the current acc_provision config. Working on it.


provision/acc_provision/templates/provision-config.yaml
```diff
@@ -68,6 +68,37 @@ net_config:
...
+                         # blocks, as well as any cluster IPs within them
+      - cidr: 192.168.8.0/22 # ServiceClusterIPBlock represents a single allowed ClusterIP
+                             # CIDR block
+    service_external_ips: # ServiceExternalIPs are the CIDR blocks for Kubernetes
+                          # Service External IPs. Kubernetes Service ExternalIPs will only be
+                          # advertised if they are within one of these blocks.
+      - cidr: 192.168.3.0/24 # ServiceExternalIPBlock represents a single allowed
+                             # External IP CIDR block.
+  bgp_peer_config:
+    name: 192-168-2-201
```
snaiksat, line 95: name is derived not static, its not part of the input

anmol372 (commented):
LGTM

provision/acc_provision/acc_provision.py
```diff
@@ -970,6 +997,56 @@ def generate_cert(username, cert_file, key_file):
             key_data = keyp.read()
     return key_data, cert_data, reused
 
+def generate_cko_cert(username, cert_file, key_file):
+    reused = False
+    if not exists(cert_file) or not exists(key_file):
+        info("Generating certs for cko")
```
snaiksat, line 1003: If the only reason we are creating a copy of this function is the above message, we can update the existing function with the logic to output the relevant message based on whether its for "kubernetes controller" or "cko". Also, rather than saying "cko", we can perhaps say "network-operator".

snaiksat:
Thanks @tanyatukade for the progress so far. How is the cleanup of the APIC configuration happening? I think we should be implementing the clean-up step as and when we implement the corresponding config add step.

tanyatukade:
> Thanks @tanyatukade for the progress so far. How is the cleanup of the APIC configuration happening? I think we should be implementing the clean-up step as and when we implement the corresponding config add step.

Working on the unprovision part.

provision/acc_provision/acc_provision.py
tanyatukade: Addressed

provision/acc_provision/acc_provision.py
```diff
@@ -1171,7 +1283,11 @@ def generate_kube_yaml(config, operator_output, operator_tar, operator_cr_output
 
 
 def generate_apic_config(flavor_opts, config, prov_apic, apic_file):
+    apic=None
+    if prov_apic is not None:
+        apic = get_apic(config)
     configurator = ApicKubeConfig(config)
+    ApicKubeConfig.init_apic(config, apic)
```
snaiksat, line 1290: We are calling "get_apic()" in L1306 (and you are already initializing apic to None in the ApicKubeConfig), so instead of doing all of the above we just need to call:

    ApicKubeConfig.init_apic(config, apic)

after L1306.

provision/acc_provision/acc_provision.py
```diff
@@ -208,6 +208,9 @@ def config_default():
         "service_mesh_config": {
             "enable": False,
         },
+        "metallb_config": {
```
shastrinator, line 211: As discussed, rename to load_balancer_config: {
"enable": False,
"lb_type": "metallb",
}

shastrinator (changes requested):

provision/acc_provision/acc_provision.py
```diff
@@ -1959,7 +1970,7 @@ def provision(args, apic_file, no_random):
             gen = globals()[gen]
         netop_output_file = args.output
         print("generating network operator output file")
-        gen(config, netop_output_file)
+        gen(args, config, netop_output_file)
```
snaiksat, line 1973: instead of passing the entire args, you can pass: config["registry"]["version"]

provision/acc_provision/acc_provision.py
```diff
@@ -1329,7 +1329,7 @@ def generate_cko_calico_yaml(config, network_operator_output):
             fh.write(network_operator_yaml)
 
 
-def generate_cko_aci_yaml(config, network_operator_output):
+def generate_cko_aci_yaml(args, config, network_operator_output):
```
snaiksat, line 1332: args -> version

and just assign that to netopConfig["aci_config"]["cni_flavor_version"], no conditional processing is required

provision/acc_provision/templates/netop-manifest.yaml
```diff
@@ -0,0 +1,1609 @@
...
+                    state:
+                      type: string
+                    status:
+                      type: string
+                    type:
+                      type: string
+                  type: object
+                description: Single String String
+                type: object
+              desiredIntent:
```
anmol372, line 253: please update to `deployedIntent`

PR #580, opened 2021-11-23
image_pull_secret check

### LINKED ISSUE #579: acc-provision: validate image_pull_secret
The image pull secret is supposed to be the name of the secret. 
However inserting the OpenShift Auth yaml code will result in a crash:

For example if you set:
 `image_pull_secret: {"auths": {"openshift-mirror.local:5000": {"auth": "redacted","email": "mail@mail.com"}}}`

The generator inside `gen_file_list` will crash as the yaml file is not valid anymore.
https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/acc_provision.py#L999

We should validate the config file for this possible mistake, the image_pull_secret is just the name of the secret and we should check for that

camrossi:
Fix #579

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -1363,7 +1363,14 @@ def check_overlapping_subnets(config):
         if out:
             return False
     return True
-
+def check_image_pull_secret(config):
+    # Check if the image_pull_secret is valid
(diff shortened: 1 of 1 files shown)


snaiksat:
#582 covers this and is merged.

PR #582, opened 2021-12-02
To validate the image_pull_secret in config file

vlella:
fix https://github.com/noironetworks/acc-provision/issues/579

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -1422,6 +1422,16 @@ def check_overlapping_subnets(config):
     return True
 
 
+def check_image_pull_secret(config):
+    # Check if the image_pull_secret is valid
+    image_pull_secret = config["registry"]["image_pull_secret"]
+    # This is the regex used by kubectl to validate objects names
+    pattern = "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+    if not re.fullmatch(pattern, str(image_pull_secret)):
+        return False
+    return True
+
+
 def provision(args, apic_file, no_random):
     config_file = args.config
     output_file = args.output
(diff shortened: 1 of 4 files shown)


snaiksat:
@vlella thanks, can you please update the commit title and message?
Let's try to follow these guidelines:
https://chris.beams.io/posts/git-commit/
https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

provision/acc_provision/acc_provision.py
```diff
@@ -1508,6 +1518,12 @@ def provision(args, apic_file, no_random):
         err("overlapping subnets found in configuration input file")
         return False
 
+    # Verify that image_pull_secret is a valid K8s secret name and not a YAML string
+    if "registry" in config.keys() and "image_pull_secret" in config["registry"]:
+        if not check_image_pull_secret(config):
+            err("The image_pull_secret is Invalid. It is supposed to be name of the k8s secret and it shouldn't have characters like =,+,!,~,@,& etc")
```
snaiksat, line 1524: Rephrase this to say:
Invalid image_pull_secret value, it must be a valid DNS subdomain name.

provision/acc_provision/acc_provision.py
```diff
@@ -1365,6 +1365,16 @@ def check_overlapping_subnets(config):
     return True
 
 
+def check_image_pull_secret(config):
+    # Check if the image_pull_secret is valid
+    image_pull_secret = config["registry"]["image_pull_secret"]
+    # This is the regex used by kubectl to validate objects names
+    pattern = "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
```
snaiksat, line 1372: Try to compare this with other examples that people might have posted for validating DNS subdomain names, and accordingly lets add a UT which validates, both, the positive and the negative cases.

snaiksat (changes requested):

provision/acc_provision/acc_provision.py
vlella: Yes updated as requested. Thanks for the input @snaiksat

provision/acc_provision/acc_provision.py
vlella: okay let me understand how to write a UT for validating DNS subdomain names then may be similarly try to mimic the work for our use case - image_pull_secret 

I guess the understanding is to add UT here - https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/test_main.py
where 67 unit tests are already written for other work.
@snaiksat

snaiksat: To clarify, what I meant is compare the regex pattern we have used with the one others have used for DNS subdomain names to ensure we are not missing anything.

vlella: okay makes sense, thanks for clarifying! @snaiksat 

Regarding adding UT 
(67 existing ones + adding 1 new one for image_pull_secret) 
I guess with respect to this the understanding is right from my end?

snaiksat (approved):

snaiksat (approved):

PR #591, opened 2022-01-05
ACI-CNI: OCP 4.8 4.9 flavor on openstack

abhis2112:


provision/testdata/list_flavors.stdout.txt
```diff
@@ -1,5 +1,7 @@
 INFO: Available configuration flavors:
 INFO: openshift-4.8-baremetal:	Red Hat OpenShift Container Platform 4.8 on Baremetal
+INFO: openshift-4.9-openstack:	Red Hat OpenShift Container Platform 4.9 on OpenStack
+INFO: openshift-4.8-openstack:	Red Hat OpenShift Container Platform 4.8 on OpenStack
```
snaiksat, line 4: The ordering is broken here. It seems that we had earlier followed the convention of grouping all the releases of a given integration together (meaning all BMs together, all OpenStacks together, etc.). I believe this was not followed correctly during the BM flavor change, so lets fix it now.

provision/testdata/flavor_openshift_49_openstack_tar/cluster-network-06-PriorityClass-acicni-priority.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: scheduling.k8s.io/v1beta1
```
snaiksat, line 1: Since 4.9 uses k8s 1.22, the above needs to be fixed in the template (only for 4.9), please see:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#priorityclass-v122

snaiksat (changes requested):

provision/testdata/list_flavors.stdout.txt
abhis2112: sure sumit

provision/testdata/flavor_openshift_49_openstack_tar/cluster-network-06-PriorityClass-acicni-priority.yaml
abhis2112: make sense yes. Following it

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -17,7 +17,11 @@ metadata:
 ---
 {% endif %}
 {% if config.kube_config.use_acicni_priority_class %}
+{% if config.flavor != "openshift-4.9-openstack" %}
```
snaiksat, line 20: We need to do this in a way such that it works for 4.9 and above. You are probably better off checking for equality match with flavors prior to 4.9.

snaiksat:
We need an update to the fix-testdata script in the support repo.

snaiksat:
There is also a pep8 failure.

provision/acc_provision/templates/aci-containers.yaml
abhis2112: sure. good point. Let me correct this

snaiksat (approved):

PR #593, opened 2022-02-03
Added device_delete_timeout in controller-config

akhilamohanan:
Added a configurable variable in controller-config of
aci-containers-config configmap which is the period after which an old
opflex device will be deleted once the ofplex device is flagged for
delete. If the value is not given in config, its value will be set to
1800s


provision/acc_provision/acc_provision.py
```diff
@@ -160,6 +160,7 @@ def config_default():
             "kube_default_provide_kube_api": False,
             "disable_node_subnet_creation": False,
             "preexisting_kube_bd": None,
+            "device_delete_timeout": None,
```
snaiksat, line 163: Changes look good but I would like to suggest a minor change - can we rename the config to opflex_device_delete_timeout

akhilamohanan: addressed

snaiksat (approved):

PR #597, opened 2022-03-07
fix ansible-operator port conflicts

mridulgain-oc:
addressing: https://github.com/noironetworks/support/issues/1756
We can schedule the ansible-operator in specific node by labeling i with `allowed-operator=aci` so that it doesn't conflict with customers ansible operator.

### DIFF
--- a/provision/acc_provision/templates/aci-operators.yaml
+++ b/provision/acc_provision/templates/aci-operators.yaml
@@ -182,6 +182,16 @@ spec:
         name: aci-containers-operator
         network-plugin: aci-containers
     spec:
+      affinity:
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - preference:
+              matchExpressions:
+              - key: preferred-node
+                operator: In
+                values:
+                - aci-containers-operator-2577247291
+            weight: 1
       containers:
       - image: {{ config.registry.image_prefix }}/aci-containers-operator:{{ config.registry.aci_containers_operator_version }}
         imagePullPolicy: {{ config.kube_config.image_pull_policy }}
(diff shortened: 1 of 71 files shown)


snaiksat:
Thanks @mridulgain-oc . Update to the UTs?

mridulgain-oc:
UTs are updated accordingly

abhis2112:
Couple of issues that I see here:
1. The label on the node has to be manually provided after install - kubectl label nodes <node-name> <label-key>=<label-value>. Ideally, it should be cluster-ready and we don't add it afterward. And to automate it in aci_deployment.yaml, we need to know the node-name beforehand.
2. The above solution can restrict the aci-containers-operator to be on a particular node. But we are not controlling the scheduling of customers ansible-operator. So, if no label/node-affinity matched - the pod can be scheduled on any of the node, and can be placed onto the same node as aci ansible-operator. 

Possible-Soln - As suggested by @snaiksat  We can provide unique healthz ports different from customer ansible-operator to clear collision

chandra-k:
> Couple of issues that I see here:
> 
> 1. The label on the node has to be manually provided after install - kubectl label nodes  =. Ideally, it should be cluster-ready and we don't add it afterward. And to automate it in aci_deployment.yaml, we need to know the node-name beforehand.
> 2. The above solution can restrict the aci-containers-operator to be on a particular node. But we are not controlling the scheduling of customers ansible-operator. So, if no label/node-affinity matched - the pod can be scheduled on any of the node, and can be placed onto the same node as aci ansible-operator.
> 
> Possible-Soln - As suggested by @snaiksat We can provide unique healthz ports different from customer ansible-operator to clear collision
Abhishek, We couldnt figure a way to change the ports ansible listens on. Can you please tell us how to change the ports?

abhis2112:
@chandra-k , Sure let me check and get back to you

snaiksat (approved):

PR #599, opened 2022-03-15
expose ApicRefreshTickerAdjust as part of acc-provision input config

mridulgain-oc:
Issue: https://github.com/noironetworks/support/issues/1759

this patch will expose *ApicRefreshTickerAdjust* as part of acc-provision input config.
It denotes how early (seconds) the subscriptions needs to be refreshed than actual subscription refresh-timeout.
- min: 1
- max: 65535
- default: 150


snaiksat:
Thanks @mridulgain , can you please update the overrides UT?

mridulgain-oc:
Thanks for pointing that out. I've updated the override related UTs.

snaiksat (approved):

snaiksat:
Thanks @mridulgain , please backport.

PR #612, opened 2022-04-12
Add size filed in logging

rupeshsahuoc:
The input takes in bytes. The size field is the limit of file size we
want to generate in cluster report.
There is no default value for size filed


akhilamohanan:
Handled in https://github.com/noironetworks/acc-provision/pull/625

PR #613, opened 2022-04-12
Add size filed in logging

rupeshsahuoc:
The input takes in bytes. The size field is the limit of file size we
want to generate in cluster report.
There is no default value for size filed

(cherry picked from commit 441fd05c55a093cae3886a981469be380dd24faa)


akhilamohanan:
Handled in https://github.com/noironetworks/acc-provision/pull/626

Issue #614, opened 2022-04-13
Current version of acc-provision does not work with Python 3.10


mpsOxygen:
This is do to how the Mapping function from collections is used:
>>> from collections import Mapping
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'Mapping' from 'collections' (/usr/lib64/python3.10/collections/__init__.py)

This is a deprecated method of using Mapping and should be migrate to the new method:
<stdin>:1: DeprecationWarning: Using or importing the ABCs
  from 'collections' instead of from 'collections.abc' is deprecated
  since Python 3.3, and in 3.10 it will stop working

tomflynn:
Closed by #622

PR #615, opened 2022-04-13
Update apic_provision.py to work with python 3.10

mpsOxygen:
This solves Issue #614 
[](https://github.com/noironetworks/acc-provision/issues/614)

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -1,4 +1,5 @@
 from __future__ import print_function, unicode_literals
+from collections.abc import Mapping
 
 import collections
 import json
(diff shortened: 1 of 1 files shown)


vlella:
I guess we better have -
try:
    from collections.abc import Mapping
except ImportError:
    from collections import Mapping 
instead of only -
from collections.abc import Mapping

as that would take care of 3.10 version which needs abc and if 3.10 isn't the case other 3.x versions and 2 versions are also handled properly. Let me know your thoughts.

@mpsOxygen @snaiksat

snaiksat:
Thanks @mpsOxygen for the PR. I agree with @vlella comment.

mpsOxygen:
I agree as well. Should I change the PR or did he make a new PR and I can close this one?

vlella:
@mpsOxygen I didn't make any new PR for now. Whatever is convenient for you we can proceed with that.
Thanks!

mpsOxygen:
Ok, I'll just make a new one as I can't edit this one.

PR #640, opened 2022-07-15
Added new RKE-1.3.13 flavor

LavanyaVaddavalli:


provision/testdata/list_flavors.stdout.txt
```diff
@@ -19,3 +19,4 @@ INFO: kubernetes-1.22:	Kubernetes 1.22
 INFO: kubernetes-1.21:	Kubernetes 1.21
 INFO: kubernetes-1.20:	Kubernetes 1.20
 INFO: RKE-1.2.3:	Rancher Kubernetes Engine min version 1.2.3
+INFO: RKE-1.3.13:	Rancher Kubernetes Engine min version 1.3.13
```
snaiksat, line 22: the latest version should be above the older versions (please see how its being done for other flavors)

provision/acc_provision/flavors.yaml
```diff
@@ -2996,6 +2996,62 @@ flavors:
...
+          type: Kubernetes
+          injected_cluster_type: RKE
+          injected_cluster_provider: Rancher
+      istio_config:
+        install_istio: False
+      multus:
+        disable: False
+    hidden: False
+    status: null
+    order: 39
```
PariyaAshok, line 3054: order should be 40 here, all remaining flavors gets incremented  by +1,  please make sure that  we don't have duplicate order for flavors.

provision/testdata/list_flavors.stdout.txt
PariyaAshok: addressed

LavanyaVaddavalli: addressed

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -82,6 +82,7 @@ registry:
 #kube_config:
   # ovs_memory_limit: "20Gi"            # override if needed, default is "1Gi"
   # reboot_opflex_with_ovs: "false"     # override if needed, default is "true"
+  # use_cluster_role: False             # default is True
```
snaiksat, line 85: I believe this is configuration is RKE flavor specific? If so we should clearly mention it here. Ideally we should have put this in a separate section for rke, or make a sub-section under kube_config for rke and put this config inside that.

LavanyaVaddavalli: It is not rke flavor specific variable.  This variable is already there in acc-provision https://github.com/noironetworks/acc-provision/blob/70c92cf2306d9400d33954cbef6f5e3a01beda3e/provision/acc_provision/acc_provision.py#L215 

We defined it in the same way in rke as well. It is used as check to include the cluster role manifest, similar to acc_provision. (https://github.com/noironetworks/kontainer-driver-metadata/blob/bc7efbf9353aed0625613d385b1193e7d190f957/rke/templates/aci-v5.2.3.go#L1731)

In the new aci-cni 5.2.3.2 patch we made for rke, we added 11 new variables. If the user wants to set the value for any of these variables, he has to set the variable in acc_provision_input.yaml  (which comes from this provision-config template). And when used with new rke flavor these variables are then added to network provider section.

All the other 10 variables are there commented in this file already.  Since the user has to be aware of these variables, I added this variable as well.

snaiksat: If it's not RKE-specific, we should not be adding making any changes in this patch to add it to the sample file. This configuration was used only in the Docker UCP and as such its not user facing (hence it was not present in this sample file in the first place).

PariyaAshok: addressed in e7c21c9f3ab422c87890ebeab8d13f9b0b827bfd

LavanyaVaddavalli: addressed in e7c21c9f3ab422c87890ebeab8d13f9b0b827bfd

snaiksat (approved):

PR #650, opened 2022-08-02
Remove nat_outgoing and encap from calico input file

tanyatukade:


shastrinator (approved):
Calico's defaults are the same as well. You shouldn't have to set anything ideally. But its ok to be explicit.

snaiksat:
@tanyatukade please backport to kmr2

PR #652, opened 2022-08-03
Added CONFORMANCE_TESTS env in controller yaml

akhilamohanan:
CONFORMANCE_TESTS env should be set to true when we run conformance
tests. Default value will be false


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -2518,6 +2518,8 @@ spec:
               value: "{{ config.kube_config.snat_operator.rdconfig_name }}"
             - name: SYSTEM_NAMESPACE
               value: "{{ config.kube_config.system_namespace}}"
+            - name: CONFORMANCE_TESTS
```
snaiksat, line 2521: Per comment on the aci-containers PR, can we create a new input config under "kube_config" section and call it "wait_for_service_ep_readiness" with default set to True. This will make it easier to set it when running the tests as opposed to having to update the aci-containers deployment spec directly.

akhilamohanan: handled in https://github.com/noironetworks/acc-provision/pull/655

PR #659, opened 2022-08-08
Add new cluster_l3out for calico external access

tanyatukade:
- [x] Configure new l3out with vrf , l3_domain, phys_domain and vlan_pool
- [x] Update the user configured l3out with the contract
- [x] Update aci_as_number derived from the APIC


provision/acc_provision/apic_provision.py
```diff
@@ -228,6 +240,32 @@ def get_phys_dom(self, domain):
         path = "/api/mo/uni/phys-%s.json" % domain
         return self.get_path(path)
 
+    def check_aep_association_to_phys_dom(self, phy_dom, aep):
```
shastrinator, line 243: This is something we can associate.. instead of checking and returning error

provision/acc_provision/apic_provision.py
```diff
@@ -228,6 +240,32 @@ def get_phys_dom(self, domain):
...
+        res = False
+        aep_dn = "uni/infra/attentp-%s" % aep
+        try:
+            tDn = self.get_path(path)["infraRtDomP"]["attributes"]["tDn"]
+            res = (tDn == aep_dn)
+        except Exception as e:
+            err("Error in getting configured Physical Domain for %s: %s" % (phy_dom, str(e)))
+        return res
+
+    def check_aep_association_to_l3_dom(self, l3_dom, aep):
```
shastrinator, line 254: This is something we can associate.. instead of checking and returning error

provision/acc_provision/apic_provision.py
tanyatukade: Sure, I have updated.

provision/acc_provision/acc_provision.py
```diff
@@ -1029,24 +1050,26 @@ def config_validate_preexisting(config, prov_apic):
...
-            err("Physical Domain %s not created on the APIC. Please create the Physical Domain and try again" % physical_domain_name)
-            return False
         aep = apic.get_aep(aep_name)
         if aep is None:
             err("AEP %s not created on the APIC. Please create the AEP and try again" % aep_name)
             return False
+        check_local_asn = apic.get_local_asn()
+        aci_as_number = config["aci_config"]["cluster_l3out"]["bgp"]["peering"]["aci_as_number"]
+        if str(aci_as_number) != check_local_asn:
+            err("aci_as_number %s provided in the input file does not match the BGP route reflector asn %s on the APIC" %
```
shastrinator, line 1070: Just add the following to the comment. "This check is only made if -a option is used. So ensure the calico manifests that are generated from this step with -a are used instead of previous ones you may have had"

provision/acc_provision/acc_provision.py
```diff
@@ -363,7 +365,19 @@ def normalize_cidr(cidr):
...
 def config_adjust(args, config, prov_apic, no_random):
-    system_id = config["aci_config"]["system_id"]
+    if is_calico_flavor(config["flavor"]):
+        l3out_name = ""
+        if not config["aci_config"]["cluster_l3out"].get("name"):
+            vlan_id = config["aci_config"]["cluster_l3out"]["svi"]["vlan_id"]
+            l3out_name = "calico-l3out-vlan-%s" % vlan_id
+            config["aci_config"]["cluster_l3out"]["name"] = l3out_name
+        l3out_name = config["aci_config"]["cluster_l3out"]["name"]
+        system_id = "calico-%s" % l3out_name if "calico" not in l3out_name else l3out_name
```
shastrinator, line 375: this should be config["flavor"]-%s and so on

provision/acc_provision/acc_provision.py
```diff
@@ -363,7 +365,19 @@ def normalize_cidr(cidr):
 
 
 def config_adjust(args, config, prov_apic, no_random):
-    system_id = config["aci_config"]["system_id"]
+    if is_calico_flavor(config["flavor"]):
```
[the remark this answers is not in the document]
snaiksat: I like this idea  but let's take this up at the time that we want to do other CNIs.

provision/acc_provision/acc_provision.py
snaiksat: We want the "calico" prefix to be present regardless of the specific calico flavor (note that we will will have multiple calico flavors depending on the tigera operator version  and/or if the flavor is to be consumed by CKO).

shastrinator: those names could all be required to have calico in them.. but yeah functionality wise looks good. This was a nit.

provision/acc_provision/acc_provision.py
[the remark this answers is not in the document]
shastrinator: ok

snaiksat (approved):

PR #675, opened 2022-08-30
Add nettools manifests

rupeshsahuoc:


provision/testdata/flavor_cko_calico-3.23.2.inp.yaml
```diff
@@ -23,6 +23,28 @@ aci_config:
...
 
+connectivity_checker:
+ error_pods_reporting_enable: true
+ reachability_test_enable:
+  periodic_sync:
+   enable_updates: true
+   interval: 120
+  external_url:
+   url: 'google.com'
+  proxy: ""
```
tanyatukade, line 34: We dont need to add another proxy field. The "cko_proxy_config" section already has the option to set the proxy. Please use the same field.

tanyatukade (changes requested):

provision/testdata/flavor_cko_calico-3.23.2.inp.yaml
rupeshsahuoc: addressed

provision/acc_provision/templates/connectivity-checker-cr.yaml
```diff
@@ -0,0 +1,94 @@
...
+ name: test-conn-check
+ namespace: nettools
+spec:
+ reachability_test_enable:
+  periodic_sync:
+   enable_updates: {{ config.connectivity_checker.reachability_test_enable.periodic_sync.enable_updates }}
+   interval: {{ config.connectivity_checker.reachability_test_enable.periodic_sync.interval }}
+  external_url:
+   url: {{ config.connectivity_checker.reachability_test_enable.external_url.url }}
+  proxy:
```
tanyatukade, line 13: We need to first check if the cko_proxy_config field is present in the input file. {% if config.cko_proxy_config %}
Only then set the proxies.

rupeshsahuoc: addressed

provision/acc_provision/templates/aci-installer-cr.yaml
```diff
@@ -5,9 +5,13 @@ metadata:
   name: installer-aci-cni
 spec:
   config:
+    connectivityCheck:
```
tanyatukade, line 8: This entire "connectivityCheck" section should only exist if config.connectivity_checker field exists in the input file. Since this is optional to set by the user. No "else" required

provision/acc_provision/templates/calico-installer-cr.yaml
```diff
@@ -5,6 +5,10 @@ metadata:
   name: installer-calico-cni
 spec:
   config:
+    connectivityCheck:
```
tanyatukade, line 8: This entire "connectivityCheck" section should only exist if config.connectivity_checker field exists in the input file. Since this is optional to set by the user. No "else" required

tanyatukade (changes requested):

provision/acc_provision/templates/calico-installer-cr.yaml
rupeshsahuoc: addressed

provision/acc_provision/templates/aci-installer-cr.yaml
rupeshsahuoc: addressed

tanyatukade (approved):

provision/acc_provision/templates/nettools-manifests.yaml
```diff
@@ -0,0 +1,319 @@
...
+                        type: boolean
+                      interval:
+                        type: integer
+                        default: 120
+                  external_url:
+                    type: object
+                    properties:
+                      url:
+                        type: string
+                        default: 'google.com'
```
snaiksat, line 50: Can we check which URL is used by the upstream kubernetes e2e conformance tests?

provision/acc_provision/acc_provision.py
```diff
@@ -1447,12 +1461,25 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
             new_deployment_file = temp
 
         if args.cko:
+            netopConfig = dict(config)
+            if "connectivity_checker" in netopConfig.keys():
```
snaiksat, line 1465: this code block is repeated, would it make sense to factor it out?

rupeshsahuoc: addressed

provision/acc_provision/templates/nettools-manifests.yaml
rupeshsahuoc: kubernetes e2e conformance tests also use same URL google.com https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/networking.go#L102

provision/acc_provision/acc_provision.py
tanyatukade: In the Installer CR we already have this check "{% if config.connectivity_checker %}", so nettools will only get populated if the input file has that config. Hence, we dont need the above check again.

tanyatukade:
Please resolve conflicts.

provision/acc_provision/acc_provision.py
```diff
@@ -1345,14 +1361,17 @@ def generate_calico_deployment_files(args, config, network_operator_output):
         custom_resources_calicoctl_yaml = calico_bgp_config_output + bgp_peer + bgp_node
 
         if args.cko:
+            netopConfig = dict(config)
+            if "connectivity_checker" in netopConfig.keys():
```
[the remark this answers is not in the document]
rupeshsahuoc: If we remove this check here, then all cases (with cko arg and without connectivity_check) it will try to generate nettools deployment files which will fail because of nettools config does not exist. So we need this check here.

provision/acc_provision/acc_provision.py
rupeshsahuoc: If we remove this check here, then all time (with cko arg and without connectivity_check) it will try to generate nettools deployment files which will fail because of nettools config does not exist. So we need this check here.

tanyatukade (approved):

PR #677, opened 2022-09-01
Retrieval of vlan_pool on OpenShift on ESX case

vlella:


vlella:
This PR is not required anymore.

PR #679, opened 2022-09-01
adding dpu config to acc-provision

Bhavanaashok33:


snaiksat:
Thanks @Bhavanaashok33 . I am trying to understand in this patch why the testdata for non-dpu related UTs are also getting  impacted. I see that in a couple of places there is a change in white spaces for config in the template which is unrelated to the dpu config. While those changes themselves look good to me, to ensure that we are not introducing any regression, it might perhaps be better to reset those changes, and check if the non-dpu testdata remains unchanged.

Bhavanaashok33:
Hi Sumit, I see that unnecessary white spaces are getting introduced. Let me pull a fresh branch and apply the patch without affecting any non-dpu configs

Bhavanaashok33:
opening a new PR. closing this.

PR #682, opened 2022-09-03
Add cko calico system id

tanyatukade:
- For cko based calico flavor, user provided system-id should override the default.


provision/testdata/flavor_cko_calico-3.23.2.inp.yaml
```diff
@@ -1,4 +1,5 @@
 aci_config:
+  system_id: calico-cni
```
snaiksat, line 2: Should we not have a UTs with and without the system_id (to avoid getting into the issue we noticed while testing)?

If we do, perhaps adding the system_id would be an overrides case?

provision/testdata/flavor_calico-3.23.2.apic.txt
```diff
@@ -341,11 +341,11 @@
         }
     }
 }
-/api/mo/uni/tn-test/flt-calico-l3out-fsvi-vl-allow-all-filter.json
+/api/mo/uni/tn-test/flt-calico-l3out-fsvi-vlan-13-allow-all-filter.json
 {
     "vzFilter": {
         "attributes": {
-            "name": "calico-l3out-fsvi-vl-allow-all-filter",
+            "name": "calico-l3out-fsvi-vlan-13-allow-all-filter",
```
snaiksat, line 348: I am wondering where the fsvi is coming from? Also a bit worried that the suffix and prefix in the names seem to be too long (only the "vlan-13" part is dynamic here?).

provision/testdata/flavor_cko_calico-3.23.2.inp.yaml
tanyatukade: I have an UT in place for this, will rename it to override

provision/testdata/flavor_calico-3.23.2.apic.txt
tanyatukade: will be taking this up as a separate PR

provision/testdata/flavor_cko_calico-3.23.2.inp.yaml
tanyatukade: Added base case and override test cases for flavor calico

PR #683, opened 2022-09-06
added helm chart for aci operator and helm option to acc-provision

siva-muni:
this PR is for the task https://github.com/noironetworks/support/issues/1879. 
- tested the helm chart on rke cluster, able to bring up the aci-operator and py-tests also passed on the cluster. 
- Need to test on other cluster (k8s and openshift) mean while please do review.
- during the review please ignore below files. these are the helm chart default files
     - aci-helm/LICENSE
     - aci-helm/README.md
     - aci-helm/templates/_helpers.tpl


snaiksat:
Could we use the acc-provision command to generate helm chart package itself rather than creating the directory structure in this repo? This package can be generated using a different argument from the one you are using for generating the values. The package can be generated and pushed to a helm repo any time acc-provision is released.

aci-helm/templates/aci-operators.yaml
```diff
@@ -0,0 +1,251 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: aci-containers-operator
+subjects:
+- kind: ServiceAccount
+  name: aci-containers-operator
+  namespace: {{ default "aci-containers-system" (.Values.kubeConfig).system_namespace }}
+---
+apiVersion: {{ default "apps/v1" (.Values.kubeConfig).use_apps_api }}
+kind: Deployment
```
psolarcz, line 156: Maybe we could separate `Deployment` from `RBAC` objects?

provision/testdata/help.stdout.txt
```diff
@@ -2,7 +2,7 @@ usage: acc_provision.py [-h] [-v] [--release] [--debug] [--sample] [-c file]
                         [-o file] [-z file] [-r file] [-a] [-d] [-u name]
                         [-p pass] [-w timeout] [--list-flavors] [-f flavor]
                         [-t token] [--test-data-out file] [--skip-kafka-certs]
-                        [--upgrade] [--disable-multus disable_multus]
+                        [--upgrade] [--disable-multus disable_multus] [--helm]
```
anmol372, line 5: Could we have two flags here:
1. [--helm-values file]
2. [--helm-chart file] = a generated aci-helm.tar.gz chart (equivalent to output of `helm package`)

Or 
1. [--helm dir] which generates both files at the dir

aci-helm/README.md
```diff
@@ -0,0 +1,4 @@
+To install
+-----------
+
+helm install <helm-rel-name> aci-helm
```
snaiksat, line 4: aci-helm -> aci-cni (here and everywhere else we have used)

snaiksat:
(didnt realize my inline comments had not gotten published)

PR #694, opened 2022-10-05
Generate netop-manager manifests based on a cko-version

tanyatukade:


snaiksat:
Thanks @tanyatukade for the quick turnaround on this. We would also need to handle image tag version corresponding to the cko-versions. My suggestion would be to add a cko-versions.yaml similar to versions.yaml which we currently have, and introduce the versions and the image tags pertaining to the CKO images in that file (and then read that file for the CKO images).

snaiksat (approved):

PR #700, opened 2022-10-10
Added 1.25 k8s flavors for acc-provision

yogesh-oc:


akhilamohanan:
Rebased in https://github.com/noironetworks/acc-provision/pull/766

PR #714, opened 2022-10-13
To make opflex metrics configurable with acc-provision

vlella:
We introduced a new field named opflex_agent_prometheus to provision_input to enable/disable opflex metrics which is set default to false. We have conditional check of this field for port info and annotations to check whether it is set to True or False in order to add them to opflex-agent container and aci-containers-host DaemonSet respectively.


provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -2272,6 +2276,9 @@ spec:
             capabilities:
               add:
                 - NET_ADMIN
+          ports:
+            - name: metrics
+              containerPort: 9612
```
snaiksat, line 2281: It will be preferable to add this also conditionally

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -2272,6 +2276,9 @@ spec:
             capabilities:
               add:
                 - NET_ADMIN
+          ports:
```
anmol372, line 2279: The port can also be conditional, as it will only be pushing information if opflex metrics are enabled.

snaiksat:
@vlella could you also please update the commit message to indicate which new configuration is being introduced and what it does (you can summarize from the opflex prometheus document shared with you).

provision/testdata/enable_opflex_agent_prometheus.inp.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+aci_config:
+  system_id: kube
+  use_legacy_kube_naming_convention: True
+  cluster_tenant: demo
+  apic_hosts:
+    - 10.30.120.100
+  apic_login:
+    username: admin
+    password: noir0123
```
snaiksat, line 12: lets change this

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -1542,9 +1542,11 @@ data:
             ,"statistics" : { "mode" : "off" }
           {% endif %}
         },
+        {% if config.kube_config.opflex_agent_prometheus %}
```
snaiksat, line 1545: It seems like this check is evaluating to False when the value of opflex_agent_prometheus is False. Hence we should not be doing this check in this place but rather should be just setting the value below (either its True or False). By doing the check here we are populating the key:value below only in the True case.

snaiksat (changes requested):

provision/acc_provision/templates/aci-containers.yaml
vlella: @snaiksat got confused by the testdatafiles.. but your logic makes sense as I cross-checked it with below sample of code as well.. 

```
d = {"A":1, "B":0, "C":None}
print(d)
print(type(d))
for i,j in d.items():
    print i,j
if d["A"]:
    print d["A"]
if d["B"]:
    print d["B"]
if d["C"]:
    print d["C"]
```

I was with the assumption that if checks whether key is present or not more than checking the value but thats not the case as you can see for the above snippet of code. It checks the values is None/0 or not.

I will remove the if check here, then it should be fine I guess.. as it would set itself to true or false depending on case.

provision/testdata/enable_opflex_agent_prometheus.inp.yaml
vlella: @snaiksat thanks for the catch, will get the dummy in picture..

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -1543,7 +1543,7 @@ data:
           {% endif %}
         },
         "prometheus": {
-            "enabled": "false"
+            "enabled": {{ config.kube_config.opflex_agent_prometheus|json }}
```
snaiksat, line 1546: I believe this is being read as a string on the aci-containers side, right now its going as a boolean and its probably not going to work. Essentially your generated yamls in the UTs should not have a diff on this line where its false, and if they do, something is not right.

provision/acc_provision/templates/aci-containers.yaml
snaiksat: Please mark this comment as resolved since you have addressed it.

snaiksat (changes requested):

provision/acc_provision/templates/aci-containers.yaml
vlella: @snaiksat did modify the PR according to str vs boolean..
Just wanted to bring this up for this PR, we only need corresponding PR in support repo for fix-testdata which is already present and nothing is required to change on aci-containers side as per my understanding.

snaiksat: Yes, no change will be required on aci-containers side if we do the change here correctly (meaning that the value's type has to match what is currently expected in aci-containers).

snaiksat (approved):

PR #719, opened 2022-10-18
git config from secrets

psolarcz:
Updated netop-manager manifest and run fix-testdata.sh


snaiksat:
This file: provision/testdata/flavor_calico-3.23.2.apic.txt is for a UT which does not exist in this branch, but we expect to pick up the UT when we rebase this branch, so we will leave the file in there.

snaiksat (approved):

PR #721, opened 2022-10-19
Add rbac role to bgp secret, move secret to calico-system ns

tanyatukade:


snaiksat (approved):

provision/acc_provision/templates/calico-provision-config.yaml
```diff
@@ -42,7 +42,7 @@ net_config:
 # Configuration for calico networking
 #
 calico_config:
-  net_config:
+  net_config: {}
```
shastrinator, line 45: I dont think this change has any effect. You can test this by hand to see if the defaults apply in this case.

tanyatukade: It does have an effect in generating the manifests correctly. Earlier, in custom_resources_aci_calico.yaml, the input from net_config were taken, but with this patch, the manifests are getting generated correctly.

Earlier it was the following:
` ipPools:
    - blockSize:
      cidr: 20.2.0.0/16
      encapsulation:
      natOutgoing:
      nodeSelector:`

With this patch, it is:
` ipPools:
    - blockSize: 26
      cidr: 20.2.0.0/16
      encapsulation: None
      natOutgoing: Disabled
      nodeSelector: all()`

shastrinator: This should have been fixed in the template that if the field value is empty use aci default.

PR #723, opened 2022-10-20
generate argo spec

psolarcz:
Generate argo spec by acc-provision, it requires:

``` bash
gitops:
  base64_encoded_argo_manifests:
```

in config.yaml


tanyatukade (approved):
LGTM

tanyatukade:
I dont see the testdata files updated, Can we add this config in the test input file and generate the output test files?

tanyatukade:
> Generate argo spec by acc-provision, it requires:
> 
> ```shell
> gitops:
>   base64_encoded_argo_manifests:
> ```
> 
> in config.yaml

what should be the value of "base64_encoded_argo_manifests" key in the above config?

psolarcz:
> 

It should be empty. I had to add it because it was complaining that: `gitops` is a str and cannot add new key to it.

snaiksat (approved):

PR #731, opened 2022-10-27
update installerCR managedComponent key to bool from string

anmol372:


snaiksat (approved):

snaiksat:
Good catch @anmol372 , should have caught it in the review, we have been bitten by this type issue before!

PR #769, opened 2022-11-24
Added parameter to add delay in bpr programming

akhilamohanan:
Added service_graph_endpoint_add_delay field which can be filled if user needs to have a delay in bpr programming for any specific services. Delay in seconds and list of name and namespaces of the services should be mentioned in service_graph_endpoint_add_delay field.


snaiksat (commented):
As a follow up patch can we please exercise the overrides UT to add some more services to ensure that user added input is is being picked up correctly? Current patch is only verifying that the default is picked up.

snaiksat (approved):

PR #778, opened 2022-11-29
Generate acc-provision-cfgmap with calico flavor

shastrinator:
As part of unmanaged CKO case for calico

### DIFF
--- a/provision/acc_provision/templates/acc-provision-configmap.yaml
+++ b/provision/acc_provision/templates/acc-provision-configmap.yaml
@@ -1,3 +1,10 @@
+{% if config.user_config.aci_config.cluster_l3out %}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ config.kube_config.system_namespace }}
+---
+{% endif %}
 
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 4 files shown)


provision/acc_provision/templates/acc-provision-configmap.yaml
```diff
@@ -180,6 +220,20 @@ data:
                 {% endif %}
                 "node_subnet": {{config.net_config.node_subnet|json}},
                 "pod_subnet": {{config.net_config.pod_subnet|json}}
```
snaiksat, line 222: we seem to be missing: cluster_svc_subnet

provision/testdata/flavor_calico-3.23.2_tar/custom_resources_aci_calico.yaml
```diff
@@ -172,4 +172,113 @@ roleRef:
...
+                "system_id": "calico-l3out-fsvi-vl",
+                "apic_hosts": [
+                    "10.30.120.100"
+                ],
+                "vrf": {
+                    "name": "mykube_vrf",
+                    "tenant": "test"
+                },
+                "cluster_l3out": {
+                    "aep": "kube-cluster",
```
snaiksat, line 208: wondering if we should set the 'name" as well here: https://github.com/noironetworks/acc-provision/blob/master/provision/testdata/flavor_calico-3.23.2.inp.yaml#L8 so that it can be validated in this test

snaiksat (approved):

PR #782, opened 2022-12-02
Dualstack support

akhilamohanan:
Support both ipv4 and ipv6 for pod_subnet, node_subnet, extern_dynamic and extern_static.


snaiksat:
Thanks @akhilamohanan . Since this is a user facing change in the config, can we provide examples in the commit message of all possible ways to provide the subnets:s
1. single string (old model, either IPv4 or v6)
2. List with one subnet, either IPv4 or v6
3. List with two subnets for dual stack, one IPv4 and one IPv6
4. List with more than two subnets, and any combination of IPv4 and IPv6 (with the implication that if both v4 and v6 are present it amounts to a dual stack case)

PariyaAshok:
@rupeshsahuoc  
IPv4/IPv6 dual-stack Kubernetes is only supported for cluster version i.e. Kubernetes 1.20 or later and Similarly OCP 4.7 or later, So We need to take care this for corresponding acc-provision flavors.

provision/testdata/cloud_tar/cluster-network-03-config.yaml
```diff
@@ -6,7 +6,9 @@ metadata:
 spec:
   disableMultiNetwork: true
   clusterNetwork:
-  - cidr: "14.3.0.1/16"
+  - cidr: [
+    "14.3.0.1/16"
+]
     hostPrefix: 23
```
PariyaAshok, line 12: Below change is required for dualstack:
cidr should not be list for dualstack case, it should be another object of clusterNetwork like below.
clusterNetwork:
  - cidr: 14.3.0.1/16
    hostPrefix: 23
  - cidr: 2001:fb8:42::/56
    hostPrefix: 64

snaiksat (approved):

PR #798, opened 2023-01-12
Adding version management for the cni target_version

abhis2112:
The PR covers the version management for cni nf with kind as aci/calico :-
1. cni/aci/versions/
2. cni/calico/versions/
3. ApicKubeConfig version based on the cni
Further additions will be made for other nfs and kinds as per the requirements.


shastrinator:
Hi Abhishek, I think tagging the ApicKubeConfig function with the version and moving that under the cni is missing.

abhis2112:
Followup - After further discussion with @shastrinator . Will be doing following changes:
1. The APICKubeConfig will be moved to a separate python file and the file will be tagged according to aci target version
2. For calico will be using default version as 3.23.2 and including that inside the the templates.

New changes:
1. We need APIC version to be provided to input file in case of multiple apic version along with target version for aci. This will be a tracked as a new issue and discussed further

shastrinator (commented):
In general, idea was to not create multiple directories unless required. Between versions A and B if there is no change, keep using directory A. For greater or equal to B use directory B. This applies to CNI and APIC. For Apic, ideally add the current methods with a version number or no version number for now. When we have APIC changes then get_apic can return a new class suffixed with version.  This is a suggestion.

provision/acc_provision/acc_provision.py
```diff
@@ -1627,9 +1643,25 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
...
 def generate_apic_config(flavor_opts, config, prov_apic, apic_file):
+    if config.get("target_version"):
+        if config["target_version"] == "5.2.3.3":
+            version = "5-2-3-3"
+        elif config["target_version"] == "5.2.3.4":
+            version = "5.2.3.4"
+        elif config["target_version"] == "5.2.3.5":
+            version = "5.2.3.5"
+    else:
+        version = "5-2-3-3"
```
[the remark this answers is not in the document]
abhis2112: so this was under the impression that if no target version is provided, we will default it to a particular version with the current implementation. In this case 5.2.3.3

I will provide the rest of the changes as per the above suggestion

shastrinator:
Couple more suggestions:
For Apic class, you can have a base form without any versions and inherit the class and only override the particular method that changed.
For CNI:  we can make a wrapper function that returns a struct containing all the individual template names that are part of the rendering as also the input. This wrapper function also takes in a version. We can do the version comparison in the function and only add files for versions instead of directories..

abhis2112:
Addressed the above comments.
For Apic Class: Inherited the parent class. Currently, no override function is defined as it will be done in the future.

For CNI: Removed the particular versions directory and directly added prefixes for the corresponding files. The need for a separate function didn't seem to be required that return a dir(struct) as wherever we have called for templates, the version is suffixed while calling the template, and the same template is provided in the path.

shastrinator:
One more suggestion:
The idea of version comparison was to avoid creating new templates for every version, if there is no change in the content. Can we reduce some of the files in this way. for eg. acc-provision-configmap-5-2-3-3.yaml and acc-provision-configmap-5-2-3-4.yaml. We can use acc-provision-configmap-5-2-3-3.yaml  for version <= 5-2-3-4.  If we know a template needs to change for a new version, then we create a file with new version from that version onwards. version comparison needs to be done at individual template level.

shastrinator (approved):
LGTM

provision/acc_provision/acc_provision.py
```diff
@@ -1478,14 +1488,27 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
         "configmap", "secret", "serviceaccount",
         "daemonset", "deployment",
     ]
+
+    if config.get("target_version"):
+        if (parse_version.parse(config["target_version"]) <= parse_version.parse('5.2.3.3')):
+            ver = "5-2-3-3"
+        elif (parse_version.parse(config["target_version"]) < parse_version.parse('5.2.3.5')):
```
snaiksat, line 1495: seems like a typo here, should this not be 5.2.3.4?

provision/acc_provision/acc_provision.py
```diff
@@ -1478,14 +1488,27 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
...
+            ver = "5-2-3-4"
+        elif parse_version.parse(config["target_version"]) >= parse_version.parse('5.2.3.5'):
+            ver = "5-2-3-5"
+    else:
+        ver = "5-2-3-3"
```
snaiksat, line 1500: if this is the default, can we change this to 5.2.3.6 assuming we add 5.2.3.6 above?

provision/acc_provision/acc_provision.py
```diff
@@ -1478,14 +1488,27 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
...
         "daemonset", "deployment",
     ]
+
+    if config.get("target_version"):
+        if (parse_version.parse(config["target_version"]) <= parse_version.parse('5.2.3.3')):
+            ver = "5-2-3-3"
+        elif (parse_version.parse(config["target_version"]) < parse_version.parse('5.2.3.5')):
+            ver = "5-2-3-4"
+        elif parse_version.parse(config["target_version"]) >= parse_version.parse('5.2.3.5'):
+            ver = "5-2-3-5"
```
snaiksat, line 1498: we have not released 5.2.3.6 but we will be soon, should we proactively add it here?

provision/acc_provision/acc_provision.py
```diff
@@ -1628,9 +1651,23 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
 
 
 def generate_apic_config(flavor_opts, config, prov_apic, apic_file):
+    if config.get("target_version"):
```
snaiksat, line 1654: same comments as earlier in L1491-1500, also why can this function not be called from there instead of duplicating the logic in two places?

snaiksat (changes requested):
Thanks @abhis2112 . What I was expecting to see if a directory per release per CNI.

E.g. provision/acc_provision/templates/cni/aci/5.2.3.x or provision/acc_provision/templates/cni/calico/3.23.x
 
Currently this is flat structure, and makes it difficult to track. We have done something similar for CKO releases. Can we do that here as well?

Some other comments inline.

snaiksat:
Seems like the directory structure was discussed in the reviews, sorry for being late to the party. I dont quite understand why we didnt have the directory structure but instead preferred to prefix the release to the names. I would prefer that we use the directory structure and not change the file names. This makes things much easier to when we move from one release to another (you just have to index into the proper directory as opposed to special casing for filenames).

abhis2112:
> Thanks @abhis2112 . What I was expecting to see if a directory per release per CNI.
> 
> E.g. provision/acc_provision/templates/cni/aci/5.2.3.x or provision/acc_provision/templates/cni/calico/3.23.x
> 
> Currently this is flat structure, and makes it difficult to track. We have done something similar for CKO releases. Can we do that here as well?
> 
> Some other comments inline.

Sure, however I started with multiple directory structure and based on @shastrinator comment , I used this flatten approach as  templates/<network-function>/<kind>/ . Earlier it was what you commented templates/<network-function>/<kind>/<version>. Probably if we can come up with the common solution ?

provision/acc_provision/acc_provision.py
abhis2112: noted

provision/acc_provision/acc_provision.py
abhis2112: the inline code is different. I will check again

deploy.yaml
```diff
@@ -0,0 +1,1631 @@
+apiVersion: v1
```
snaiksat, line 1: is this file required?

snaiksat (changes requested):
There are a bunch of deploy.yaml (and one with the name deply.yaml) and a abc.txt, are all these files needed??

deploy.yaml
abhis2112: thanks @snaiksat , this must have been added by mistake. fixing

shastrinator:
We didn't complete going through the files: These templates are not likely to have any changes between versions:
ACI:
aci-operators-cr.yaml
aci-installer-cr.yaml
platform-installer-cr.yaml
Calico:
calico-bgp-config.yaml
calico-bgp-peer.yaml
calico-node.yaml
calico-installer-cr.yaml
platform-installer-cr.yaml

provision/acc_provision/acc_provision.py
```diff
@@ -1628,9 +1653,26 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
...
+        if (parse_version.parse(config["target_version"]) <= parse_version.parse('5.2.3.3')):
+            ver = "5_2_3_3"
+        elif (parse_version.parse(config["target_version"]) == parse_version.parse('5.2.3.4')):
+            ver = "5_2_3_4"
+        elif parse_version.parse(config["target_version"]) == parse_version.parse('5.2.3.5'):
+            ver = "5_2_3_5"
+        elif parse_version.parse(config["target_version"]) >= parse_version.parse('5.2.3.6'):
+            ver = "5_2_3_6"
+    else:
```
snaiksat, line 1666: I think I had asked earlier, why does this default to 5.2.3.3? Should it not default to the latest, and preferably, should this not be coming from a "default" config which we can change as desired as opposed to revisiting this code? (sorry if I am missing something basic)

abhis2112: this was default to 5.2.3.3 if no target_version is defined in input_file. If we want to default to the latest version then sure I can do that. I was having the impression to default it to the last stable version

snaiksat: Lets also pick up this value from the place where you are picking up the other default values. Also, can we not generically parse this version and construct the version name with the underscore in it? That way we just need to change the default value in the config whenever the new CNI version is released, and not code change is required here. If the user enters a version which is not available in the templates, the template lookup should error out with the appropriate message that the version is not supported.

provision/acc_provision/apic_provision.py
```diff
@@ -8680,5 +8680,25 @@ def rke_flavor_specific_handling(aci_prefix, data, ports, api_filter_prefix, rke
...
+
+
+class ApicKubeConfig_5_2_3_5(ApicKubeConfig):
+
+    ACI_PREFIX = aci_prefix
+
+
+class ApicKubeConfig_5_2_3_6(ApicKubeConfig):
+
+    ACI_PREFIX = aci_prefix
```
snaiksat, line 8700: Does a new class have to be created, or can we pass the version as an attribute to the class? I would prefer that we dont have to create a new class every time.

abhis2112: so this is as per the earlier comment where we inherit the original class and do changes only required for the current version in this class. That was the main idea behind making a new class. So that changes becomes minimum @shastrinator  could you please provide more info as per your comments

provision/acc_provision/acc_provision.py
shastrinator: there is a default_version in the config, I missed that the target_version is not passed in the config today. default_version comes from the flavor and is always set. You should use this instead of target_version.

provision/acc_provision/apic_provision.py
shastrinator: Yes, if we pass the version as an attribute, then we end up with checks for versions inside apic methods. This way we can override a method in a specific apic version, if there are no changes- then there will be those 2 lines here.

provision/acc_provision/acc_provision.py
abhis2112: ok that make sense about the default version. Let me check that as well

provision/acc_provision/apic_provision.py
shastrinator: We can additionally avoid creating a new apic class for very version by introducing a apic_class_version variable and keeping it the same for all the versions, till a new apic_class_version is needed at which time, then and future versions will use the new class.

shastrinator: Since we don't anticipate the APIC class changing actively/at all (ACI won't break backward compatibility). So we can do this on a need basis and have APIC class be a static class regardless of target versions. At the point that it does change, we can do this change.

shastrinator: Besides, there is already a get_config method that takes the apic-version as input and does apic-version specific changes.

abhis2112: Thanks @shastrinator for providing the inputs. Will be reverting multiple class changes to the original design

PR #799, opened 2023-01-20
CKO: control-cluster - Remove type field from svi

abhis2112:
The confi.yaml has type field but the code svi struct doesnot have the type field . This field while looking over acc-provision code is not doing anything with it. So removed from acc-provision and control cluster logic

### DIFF
--- a/provision/testdata/flavor_calico-3.23.2_tar/custom_resources_aci_calico.yaml
+++ b/provision/testdata/flavor_calico-3.23.2_tar/custom_resources_aci_calico.yaml
@@ -219,7 +219,6 @@ data:
                         "floating_ip": "2.100.101.100/24",
                         "mtu": 9000,
                         "secondary_ip": "2.100.101.254/24",
-                        "type": "floating",
                         "vlan_id": 13
                     }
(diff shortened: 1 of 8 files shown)


provision/testdata/flavor_calico-3.23.2_tar/custom_resources_aci_calico.yaml
```diff
@@ -196,7 +196,7 @@ data:
                 "enable_updates": false
             },
             "aci_config": {
-                "system_id": "calico-l3out-fsvi-vlan-13",
+                "system_id": "calico-cni",
```
snaiksat, line 199: why is this changing? are we sure the default type is getting set correctly to floating?

abhis2112: yes, I saw this before committing. The change I made was for only type: floating and then I ran the fix-testdata.sh but that I believe changed the default. I will try again to see why this change is coming which is unrelated

snaiksat (approved):

PR #800, opened 2023-01-23
Calico multiple vrf

akhilamohanan:


provision/acc_provision/apic_provision.py
```diff
@@ -603,25 +625,47 @@ def update(data, x):
...
+                4. Set shared route control on the internal epg under floating svi l3out
+                5. Set Shared route control on the external epg under external l3out
+                   User should do this ?
+                """
+                # different tenant, different vrf
+                update(data, self.l3out_brcp_global())  #1
+                update(data, self.cluster_l3out_cif_global())  #2
+                update(data, self.ext_epg_svc_contract_interface())   #3,4
+                update(data, self.ext_epg_int_contract_interface())   #3
+            elif cluster_l3out_vrf_details["create_vrf"] == True and cluster_l3out_vrf_details["create_tenant"] == False:
```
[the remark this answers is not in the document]
shastrinator: Yes your comment is correct. That is what we want to do in this case.

provision/acc_provision/acc_provision.py
```diff
@@ -915,6 +915,11 @@ def Raise(exception):
         else:
             extra_checks = {}
     elif is_calico_flavor(config["flavor"]):
+        if config["aci_config"]["cluster_l3out"].get("vrf") is not None:
```
shastrinator, line 918: Ok, if only cluster_l3out tenant is mentioned and not cluster_l3out vrf, then we can generate different vrf that we want to create as the tenantname_vrf. ( instead of checking-this is a change from the suggestion earlier). Of course if user mentions the vrf also under cluster_l3out, then we use it.  On the other hand, if only cluster_l3out vrf is mentioned, then we assume same tenant as external_l3out.

provision/acc_provision/apic_provision.py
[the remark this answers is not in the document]
shastrinator: Just use the global scope contract in all cases. Use the contract interface in case of multiple tenants  and contract in case of single tenant.

provision/acc_provision/apic_provision.py
```diff
@@ -2752,6 +2849,175 @@ def l3out_filter(self):
...
+                    "vzCPIf",
+                    collections.OrderedDict(
+                        [
+                            (
+                                "attributes",
+                                collections.OrderedDict(
+                                    [
+                                        (
+                                            "name",
+                                            "%s-l3out-allow-all"
```
shastrinator, line 2870: It would be better to name this to xxx-export so it is clear that this is the exported contract

provision/acc_provision/apic_provision.py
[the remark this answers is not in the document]
shastrinator: Update to his: please create the global scope contract where vrf is different only.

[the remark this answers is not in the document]
shastrinator: and contract-if should be created in cases where tenants are different AND either of the tenants is not tenant-common.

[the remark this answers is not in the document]
pkharat: Okay. Regarding scope, please confirm this @shastrinator @fwardzic 

1. when different vrfs under same/different tenant then **scope = "export-rtctrl,import-rtctrl,import-security,shared-security"**
for all subnets under "net_config" i.e  extern_dynamic, node_subnet, pod_subnet, cluster_svc_subnet

2. when same vrf then **scope = "export-rtctrl"**
for all subnets under "net_config" i.e  extern_dynamic, node_subnet, pod_subnet, cluster_svc_subnet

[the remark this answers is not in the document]
shastrinator: Yes looks good.

provision/acc_provision/apic_provision.py
```diff
@@ -603,25 +660,57 @@ def update(data, x):
...
 
             update(data, self.l3out_filter())
-            update(data, self.l3out_brcp())
-            update(data, self.ext_epg_svc())
-            update(data, self.ext_epg_int())
+
+            external_l3out_tn = self.config["aci_config"]["vrf"]["tenant"]
+            external_l3out_vrf = self.config["aci_config"]["vrf"]["name"]
+
+            if cluster_l3out_tn != external_l3out_tn and cluster_l3out_vrf != external_l3out_vrf:
```
shastrinator, line 683: If tenant is different and provider tenant is "common" , then we don't need to export the contract. So add condition external_l3out_tn != "common" here

provision/acc_provision/apic_provision.py
```diff
@@ -603,25 +660,57 @@ def update(data, x):
...
+                1. Create global scope contract under provider (vrf) tenant
+                2. Export global scope contract under consumer(cluster_l3out) tenant
+                3. Consumer tenant has the floating svi l3 out. The ext epgs under floating
+                   l3out consume the exported contract
+                """
+                update(data, self.l3out_brcp_global())
+                update(data, self.cluster_l3out_cif_global())
+                update(data, self.ext_epg_svc_contract_interface())
+                update(data, self.ext_epg_int_contract_interface())
+            elif cluster_l3out_tn == external_l3out_tn and cluster_l3out_vrf != external_l3out_vrf:
```
shastrinator, line 695: an additional case for the same block. If tenants are different but external_l3out_tn == common, fall into here.

provision/acc_provision/test_main.py
```diff
@@ -1104,6 +1103,30 @@ def test_flavor_calico_3_23_2():
...
+    run_provision(
+        "flavor_calico_3.23.2_same_tenant_diff_vrf.inp.yaml",
+        None,
+        "flavor_calico_3.23.2_same_tenant_diff_vrf_tar",
+        None,
+        "flavor_calico_3.23.2_same_tenant_diff_vrf.apic.txt",
+        overrides={"flavor": "calico-3.23.2"}
+    )
+
+
```
shastrinator, line 1129: Add case for provider tenant being common as well.

shastrinator (changes requested):
Please address common tenant changes as noted.

provision/acc_provision/apic_provision.py
pkharat: intention behind avoiding cluster_l3out / external_l3out tenant name check was to handle all the scenarios, where tenants are different, in the same way of creating contract interface. Benefit seen is to avoid multiple such checks and same if block can handle all different tenant scenarios using similar kind of configuration

pkharat: @shastrinator Addressed this one

provision/acc_provision/apic_provision.py
pkharat: @shastrinator Addressed this one

pkharat:
> Please address common tenant changes as noted.

taken care

provision/acc_provision/test_main.py
```diff
@@ -1105,6 +1104,41 @@ def test_flavor_calico_3_23_2():
...
+        "flavor_calico_3.23.2_multiple_vrf_uc3.inp.yaml",
+        None,
+        "flavor_calico_3.23.2_multiple_vrf_uc3_tar",
+        None,
+        "flavor_calico_3.23.2_multiple_vrf_uc3.apic.txt",
+        overrides={"flavor": "calico-3.23.2"}
+    )
+
+
 def test_flavor_with_cluster_svc_export_calico_3_23_2():
```
shastrinator, line 1142: You are missing the decorator @in_testdir above this line

shastrinator (approved):

PR #817, opened 2023-02-08
Added k8s 1.25 flavor

akhilamohanan:
Added support for k8s 1.25 flavor


snaiksat:
We intend to support only three k8s flavors, so as we support new flavors we should set the older flavors to hidden (we are showing the flavors all the way back to 1.20).

snaiksat (approved):

PR #854, opened 2023-03-07
Moving base folder yaml templates inside versioned cni folder

abhis2112:
The PR provides following fixes:
1. All the aci/calico related cni templates are moved to versioned folder
2. The base templates are moved to legacy folder for the record.
3. Following files are moved to templates/cko/netop_version/ folder: a. calico-installer-cr.yaml b. platform-installer-cr.yaml c. argocd.yaml
4. The cni target_version calculation is centralized.


provision/acc_provision/acc_provision.py
```diff
@@ -1611,12 +1599,12 @@ def generate_kube_yaml(args, config, operator_output, operator_tar, operator_cr_
...
             network_operator_openshift_spec_output = network_operator_openshift_spec_template.render(config=config)
 
-            network_operator_CR_template = get_jinja_template('cni/aci/' + ver + '/aci-installer-cr.yaml')
+            network_operator_CR_template = get_jinja_template(ACI_CNI_VERSIONS_PATH + '/aci-installer-cr.yaml')
             base64_encoded_cko_aci_spec = base64.b64encode(new_deployment_file.encode('ascii')).decode('ascii')
             netopConfig["aci_config"]["operator_version"] = netop_version
             netopConfig["aci_config"]["cni_flavor_version"] = config["registry"]["version"]
             netopConfig["aci_config"]["base64_encoded_cko_aci_spec"] = base64_encoded_cko_aci_spec
-            network_operator_platform_CR_template = get_jinja_template('cni/aci/' + ver + '/platform-installer-cr.yaml')
+            network_operator_platform_CR_template = get_jinja_template('cko/' + netop_version + '/platform-installer-cr.yaml')
```
snaiksat, line 1607: Why are we not following the pattern of computing the path once and assigning it to a variable for CKO as well?

provision/acc_provision/acc_provision.py
```diff
@@ -2127,8 +2115,21 @@ def provision(args, apic_file, no_random):
     config["aci_config"]["sync_login"]["cert_data"] = cert_data
     config["aci_config"]["sync_login"]["cert_reused"] = reused
 
+    if not is_calico_flavor(config["flavor"]):
+        global ACI_CNI_VERSIONS_PATH
```
snaiksat, line 2119: Use of global is generally discouraged. One option would have been to set these paths as key/values in the "config" dict. (in case we want to use these paths even outside this module). If it seems like an overkill, we can stay with current pattern.

provision/acc_provision/acc_provision.py
abhis2112: so I thought of maintaining the previous implementation that existed in the code. Here - https://github.com/noironetworks/acc-provision/pull/854/files/0bf5d4ca13bcfbb798758310965a902449369425#diff-d18edd87a3bcaa21be667d28554472484dbc49b6c8839bfddec4c68545f9c559R1481 . Should we change this? and will change everywhere cko/ + netop_version is used?

provision/acc_provision/acc_provision.py
abhis2112: Ohkk Sumit, Let me quickly have it, should not be a big change

provision/acc_provision/acc_provision.py
snaiksat: I think we should...

provision/acc_provision/acc_provision.py
snaiksat: we can skip this or take it as follow up (but I would say we should do the other change for cko + netop_version)

provision/acc_provision/acc_provision.py
abhis2112: sure working on both comments

anmol372 (approved):

snaiksat (approved):

PR #865, opened 2023-03-23
Added RKE2 flavor for kubernetes version 1.24

LavanyaVaddavalli:
The RKE2 flavor is similar to kubernetes flavors. 
But added two extra flags use_cnideploy_initcontainer and allow_pods_external_access, since we need these two variables set in case of RKE2.
And also added the vmm domain similar to RKE, so that it is visible under 'Rancher' VMM domain.

Wrote a single testcase to test this flavor

Testing is Pending on a ACI 5.x setup

### DIFF
--- a/provision/acc_provision/acc_provision.py
+++ b/provision/acc_provision/acc_provision.py
@@ -290,6 +290,10 @@ def config_default():
         "lb_config": {
             "lb_type": "metallb",
         },
+        "rke2_config": {
+            "logging_namespace": "cattle-logging",
+            "monitoring_namespace": "cattle-prometheus",
+        },
     }
     return default_config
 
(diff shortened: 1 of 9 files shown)


LavanyaVaddavalli:
> Are we missing the cattle-system namespaces annotation that was part of RKE?

@shastrinator cattle-system namespace will be there in RKE2 clusters as well with rancher agent pods. I am not aware of any annotations in RKE. Can you give me more context on what you are referring to?

shastrinator:
I see you have added  allow_kube_api_default_epg:true, that should work instead of the annotation on cattle-system namespace. Also all the monitoring related contracts - they need to be in RKE2 too right(https://github.com/noironetworks/acc-provision/blob/941fbff0dfd53e1376b10337b20f697cd7c1318d/provision/acc_provision/flavors.yaml#L3936)

snaiksat:
If we need to handle the cattle namespace(s), we might need to do something to what we do here for the OpenShift namspaces:
https://github.com/noironetworks/acc-provision/blob/master/provision/acc_provision/acc_provision.py#L595-L600

snaiksat (approved):

PR #897, opened 2023-05-05
To handle pre_existing_tenant scenario as unprovisioning should not d…

vlella:
…elete a tenant it didn't create. If tenant is manually created on the APIC before provisioning have tenant name in acc provision input file under aci_config section.


snaiksat:
Thanks @vlella . I don't see any UTs changing. Should this change not exercise some UT? If a UT is missing we should add it. We should expect to see that the APIC yaml is different for the pre-existing case.

snaiksat (approved):

PR #912, opened 2023-05-18
add a newline before rendering aci-operators-configmap.yaml

anmol372:


anmol372:
unsure why adding a line is required now. this had been working for a long time previously

snaiksat (approved):

PR #919, opened 2023-05-23
Add check for validating dualstack configuration

rupeshsahuoc:


snaiksat:
Thanks @rupeshsahuoc . Can you please update the commit message clearly indicating which combinations of single stack and dual stack inputs will be validated, which will pass the validation and which will fail the validation?

If multiple subnets are provided, we want don't want to incorrectly detect the intent that its dual stack (it could be multiple incorrect subnets for single stack, either v4 or v6) and then show them the message that they must provide *both* v4 and v6.

provision/acc_provision/acc_provision.py
```diff
@@ -2073,6 +2073,39 @@ def is_support_dualstack(flavor):
...
+
+def is_valid_dualstack_config(config):
+    if "net_config" not in config:
+        return False
+    if not has_contains_v6(config["net_config"]["node_subnet"]) or not has_contains_v6(config["net_config"]["pod_subnet"]):
+        return False
+
+    if has_contains_v6(config["net_config"]["extern_dynamic"]) or has_contains_v6(config["net_config"]["extern_static"]):
+        if has_contains_v6(config["net_config"]["pod_subnet"]) and has_contains_v6(config["net_config"]["node_subnet"]):
+            return True
```
PariyaAshok, line 2104: We can get/fill this info once and use, no need call func has_contains_v6 multiple time, It would be better if we can have function which check IPV4 Only case, IPV6 Only case and dual stack case, for all other case we can just raise error with appropriate message.

PariyaAshok (changes requested):

snaiksat:
All ipv4 is not mentioned as a valid use case in the examples in the commit message. ;-)

provision/acc_provision/acc_provision.py
rupeshsahuoc: addressed

provision/acc_provision/acc_provision.py
```diff
@@ -2184,6 +2230,10 @@ def provision(args, apic_file, no_random):
         err(" Dualstack feature is not supported. Please upgrade to Kubernetes version 4.21 or later or OpenShift version 4.8 or later.")
```
snaiksat, line 2230: Although not part of this patch, there is a typo here, kubernetes version should be 1.21

provision/acc_provision/acc_provision.py
```diff
@@ -2019,58 +2019,104 @@ def has_multi_subnet(config):
...
+def determine_subnet_type(subnet_info):
+    if len(subnet_info) == 0:
+        return ""
+
+    has_ipv4 = False
+    has_ipv6 = False
 
     for subnet in subnet_info:
         rtr, _ = subnet.split("/")
         ip = ipaddress.ip_address(rtr)
```
snaiksat, line 2074: what if the address malformed, this conversion will throw an exception and acc-provision will exit with a stack trace?

provision/acc_provision/acc_provision.py
```diff
@@ -2184,6 +2230,10 @@ def provision(args, apic_file, no_random):
         err(" Dualstack feature is not supported. Please upgrade to Kubernetes version 4.21 or later or OpenShift version 4.8 or later.")
         return False
 
+    if is_dualstack_config(config) and not is_valid_dualstack_config(config):
+        err("Please provide a valid Dualstack configuration with both IPv4 and IPv6 addresses for node and pod subnets.")
+        return False
```
snaiksat, line 2235: I am still not comfortable with this logic here. If the address is malformed, how would we know if the intent was to specify an ipv4 or ipv6 address? And if you cannot infer that conclusively, how can we infer that the intent was to create dual stack?

I would rather list out all the subnets we have processed indicating which ones we have recognized as ipv4, which we have recognized as ipv6, and if there is a mix of ipv4 and ipv6 subnets, indicate clearly which incorrect subnet(s) caused us to conclude that this is not a dual stack case.

snaiksat (changes requested):

snaiksat:
@rupeshsahuoc  please also add a UT to cover each of the valid and invalid scenarios you have listed in the commit message.

provision/acc_provision/acc_provision.py
rupeshsahuoc: addressed

provision/acc_provision/acc_provision.py
rupeshsahuoc: addressed

provision/acc_provision/acc_provision.py
rupeshsahuoc: addressed

provision/acc_provision/acc_provision.py
snaiksat: how? I still don't see a try/catch block...I see a check for empty string, but the ip address could be malformed for other, reasons, no?

provision/acc_provision/acc_provision.py
snaiksat: how?

rupeshsahuoc: This process involves verifying the correctness of all subnets. If any subnets are found to be malformed, an error will be thrown indicating the specific address that is malformed. Conversely, if all subnets are valid, the next step is to validate the dualstack configuration. In case any aspect of the configuration is incorrect, an error will be displayed highlighting the specific component that needs correction.

snaiksat (approved):

snaiksat:
Thanks @rupeshsahuoc  for making the changes. @PariyaAshok currently merging is blocked, awaiting your review/approval. 

@rupeshsahuoc once merged please backport to the relevant branches.

PariyaAshok (approved):

PR #922, opened 2023-06-01
Set memory request and limit in ACI-CNI containers

pkharat:


provision/acc_provision/templates/aci-containers.yaml
PariyaAshok, line 1: With this changes, aci-containers-host pod memory resource limit be 15GB(3 * 5) and for controller pod it will be 6 GB(3 * 2), Is this correct expectations???

provision/acc_provision/acc_provision.py
PariyaAshok, line 1: We are missing check like below:
https://github.com/noironetworks/acc-provision/blob/d38c31bad4e507c4cff02717874d956f37804e2c/provision/acc_provision/acc_provision.py#L648

provision/acc_provision/acc_provision.py
```diff
@@ -225,6 +225,8 @@ def config_default():
             "kubectl": "kubectl",
             "system_namespace": "aci-containers-system",
             "ovs_memory_limit": "1Gi",
+            "containers_controller_memory_limit": "3Gi",
+            "containers_host_memory_limit": "3Gi",
```
PariyaAshok, line 229: Can we provide container specific resource limits options?? also we can remove containers_ prefix in names.

pkharat: That would become too granular and will end up showing many such options in kube_confg

provision/acc_provision/templates/aci-containers.yaml
PariyaAshok: Yes, only 3 containers will be running, 
You have done changes at 5 places, we can remove resource limits other than containers(aci-containers-host ,opflex-agent  and mcast-daemon)

pkharat: Depending on scenario those 2 containers (opflex-server, lsyncd-daemon) would run and when they run are suggesting we don't need this limit for these 2 ?

provision/acc_provision/acc_provision.py
PariyaAshok: memory usage of mcast-daemon container is less as compared to aci-containers-host opflex-agent , so i think better to provide container specific options, what do you think @snaiksat ??

provision/acc_provision/acc_provision.py
pkharat: Since we are adding these variables in config_default(), looks like we don't have to explicitly add if checks. Either default one or user provided one will be selected during deep_merge(). Verified this code by adding these variables with non-default value in test with_overrides.inp.yaml file.

provision/acc_provision/acc_provision.py
pkharat: We are setting memory limit not request. Limit will not allocate 3Gi, it will make sure containers does acquire more than this.

PariyaAshok: By default requests memory is set same as limits.

name: aci-containers-openvswitch
resources:
limits:
memory: 2Gi
requests:
memory: 2Gi

snaiksat:
Good discussion, @pkharat  and @PariyaAshok (good point about request limit, I did not realize that is being set when we are setting memory limit). I am wondering if we should set the request and memory limit to 128 MB and 3 GB respectively for the aci-containers-system namespace:
https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/

provision/acc_provision/acc_provision.py
pkharat: If we set only "limit" then k8s sets "request" value same as "limit". 
https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#what-if-you-specify-a-container-s-limit-but-not-its-request
That answers why we see this behaviour. Along with limit we need to set request as well.

pkharat: Added LimitRange namespaced resource to set aci-containers pod's default memory request and limit to 128Mi and 3Gi respectively. Haven't changed existing limit parameter of openvswitch pod.

PariyaAshok: We can remove config param (https://github.com/noironetworks/acc-provision/blob/85d03ea182aebe188a6ae8bec2eb2a0132a8c67a/provision/acc_provision/templates/provision-config.yaml#L101) to avoid conflict with newer namespace wide changes.

snaiksat: i thought about it earlier but i think we can leave it, (a) for backward compat and (b) also providing the option to fine tune OVS limits

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -101,6 +101,8 @@ registry:
         #   delay: 60                   # per service delay that overrides common delay
   # add_external_subnets_to_rdconfig: True               #override if needed, default is False
   # ovs_memory_limit: "20Gi"            # override if needed, default is "1Gi"
+  # aci_containers_memory_request: "128Mi", # kind: LimitRange, memory request, override if needed, default is "128Mi"
+  # aci_containers_memory_limit": "3Gi",    # kind: LimitRange, memory limit, override if needed, default is "3Gi"
```
snaiksat, line 105: nit: typically we provide a non-default value as an example

provision/acc_provision/acc_provision.py
PariyaAshok: With this we will be setting 2 limit 1Gi(pod specific) and 3 Gi(ns wide limit), not sure if that will work.

pkharat: @PariyaAshok Pod specific values will take precedence over namespace specific values.
Okay @snaiksat, so will not touch existing ovs memory limit references for backward compatibility

provision/acc_provision/templates/provision-config.yaml
pkharat: will change it

provision/acc_provision/acc_provision.py
PariyaAshok: @snaiksat  With namespace wide limit set, we see following, With that we are allowing max memory of aci-containers-host pod to 9Gi and operator pod to 6Gi.

Age
  ---------                                    ----                                                          ------------  ----------  ---------------  -------------  ---
  aci-containers-system                        aci-containers-controller-859b879c94-7bw88                    0 (0%)        0 (0%)      128Mi (0%)       3Gi (20%)      2m40s
  aci-containers-system                        aci-containers-host-rg692                                     0 (0%)        0 (0%)      384Mi (2%)       9Gi (61%)      24s
  aci-containers-system                        aci-containers-openvswitch-4cqwd                              0 (0%)        0 (0%)      1Gi (6%)         1Gi (6%)       24s
  aci-containers-system                        aci-containers-operator-7c6b5d6f84-66m7m                      0 (0%)        0 (0%)      256Mi (1%)       6Gi (41%)      3m19s

snaiksat (approved):

snaiksat:
@pkharat thanks, please asap backport as relevant.

PR #923, opened 2023-06-02
Pinned cryptography version to 40.0.2

akhilamohanan:
cryptography 41.0.0 requires working Rust toolchain. So, pinning cryptograhy to 40.0.2

### DIFF
--- a/provision/setup.py
+++ b/provision/setup.py
@@ -40,6 +40,7 @@ setup(
         ]
     },
     install_requires=[
+          'cryptography==40.0.2',
           'requests',
           'pyyaml',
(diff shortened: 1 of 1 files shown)


snaiksat:
Thanks @akhilamohanan , we should treat this as a temporary workaround to unblock other PRs while we look for a more permanent fix. The permanent fix should not involve pinning dependencies so that we are able to pick up the latest CVE fixes . This does come with a slight risk of breakages (like in this case) when a new version of a dependency is incompatible.

snaiksat (approved):

PR #940, opened 2023-06-06
Travis - Fix python pkg "cryptography" dependency issue

pkharat:


.travis.yml
```diff
@@ -1,7 +1,7 @@
 dist: bionic
 os: linux
 language: python
-python: 3.6
+python: 3.8
```
snaiksat, line 4: Is there a reason we decided to use 3.8 versus the later verssions? Per discussion earlier, I am using 3.9 in the kmr2-5.2.7 branch.

snaiksat (approved):

.travis.yml
pkharat: @snaiksat no specific reason as we can use any python version >= 3.8. Will use python 3.9

PR #947, opened 2023-06-08
Change acicni-priority class value

pkharat:


provision/testdata/with_overrides.inp.yaml
```diff
@@ -71,6 +71,10 @@ kube_config:
       sleep_time_snat_global_info_sync: 60
   image_pull_policy: IfNotPresent
   opflex_agent_policy_retry_delay_timer: 10
+  ovs_memory_limit: "2Gi"
+  aci_containers_memory_request: 512Mi
+  aci_containers_memory_limit: 4Gi
```
snaiksat, line 76: L75-76:
Please add this as a separate change, its not related to this PR. Adding meaningful but unrelated changes is bad for several reasons (1) it makes the PR harder to understand and review (note that this this change was slid in, not even mentioned in the commit message), (2) let's say if we were to revert this commit with the intention of reverting the priority class change (hypothetically speaking), we would lose this required change as well.

pkharat: > L75-76:

Will have to revert L74-L76, as "ovs_memory_limit" is also not related to this PR.

> Please add this as a separate change, it's not related to this PR. Adding meaningful but unrelated changes is bad for several reasons (1) it makes the PR harder to understand and review (note that this change is not even mentioned in the commit message), (2) let's say if we were to revert this commit with the intention of reverting the priority class change (hypothetically speaking), we would lose this required change as well.

Commit message point "3. Update with_overrides test"(its generic in a way and did not mention all the variable names because of number) was added to bring to the attention that some extra changes are done in this test, so that it does not go unnoticed.

Will address this.

snaiksat: Understand, but when you are making an orthogonal change (which you should not be in the first place), you need to highlight. We first cut a corner by avoiding to post a new PR, then we cut another corner by not being explicit!

snaiksat:
On second thoughts, in the current PR lets not change the defaults to system-node-critical (if we do this we need to do it in a backward compatible way and requires more thought); lets only keep the change down to being able to configure the acicni priorityclass value. Apologies for the churn from the change in thought.

pkharat:
> On second thoughts, in the current PR lets not change the defaults to system-node-critical (if we do this we need to do it in a backward compatible way and requires more thought); lets only keep the change down to being able to configure the acicni priorityclass value. Apologies for the churn from the change in thought.

Okay. Will change accordingly.

PR #966, opened 2023-06-20
Add missing unit tests - kmr2

pkharat:


snaiksat:
I think we can close this for now since the testdata was not supposed to be present in the first place.

PR #968, opened 2023-06-27
Fix - set openvswitch containers memory limit/request

pkharat:


PariyaAshok (approved):

pkharat:
After discussion with Sumit, we can stay with the existing behaviour where openvswitch pod is allocated default 1Gi memory. So closing this PR

PR #973, opened 2023-06-29
Provide option to set openvswitch container memory request

pkharat:


PariyaAshok:
Can we add check to avoid  failure case when  user provided memory request value is greater than memory limit.

PariyaAshok (approved):

snaiksat (approved):

PR #977, opened 2023-07-05
Allow memory request and limit configuration per container

pkharat:


PariyaAshok:
Why can't we set request/limit for other containers like OVS container(https://github.com/noironetworks/acc-provision/blob/46e42e68a36deca6c50968356f089484b331c593/provision/acc_provision/templates/aci-containers.yaml#L2499-L2502) ?? or can we  update new change for OVS container also for sack of uniformity??

provision/acc_provision/templates/aci-containers.yaml
```diff
@@ -2202,6 +2202,21 @@ spec:
...
+              {% else %}
+              memory: "{{ config.kube_config.aci_containers_memory_limit }}"
+              {% endif %}
+            requests:
+              {% if config.kube_config.aci_containers_host_memory_request %}
+              memory: "{{ config.kube_config.aci_containers_host_memory_request }}"
+              {% else %}
+              memory: "{{ config.kube_config.aci_containers_memory_request }}"
+              {% endif %}
+          {% endif %}
```
PariyaAshok, line 2219: do we really need this additional if checks, namespace wide request and limit will automatically apply right??

pkharat:
> Why can't we set request/limit for other containers like OVS container(
We wan't to take benefit of aci_containers_memory_request, aci_containers_memory_limit in case user does not want to set per container 

>  or can we update new change for OVS container also for sack of uniformity??
Thought of changing ovs section but avoided for the same reason of not touching already tested part

provision/acc_provision/templates/aci-containers.yaml
pkharat: Memory request value is not picked from namespaced resource, it is set same as container memory limit.
(this was the issue with ovs_memory_limit and we had to explicitly add new flag ovs_memory_request)

provision/testdata/with_overrides.inp.yaml
```diff
@@ -75,6 +75,18 @@ kube_config:
...
+  aci_containers_controller_memory_limit: "5Gi"
+  aci_containers_host_memory_request: "256Mi"
+  aci_containers_host_memory_limit: "5Gi"
+  mcast_daemon_memory_request: "256Mi"
+  mcast_daemon_memory_limit: "5Gi"
+  opflex_agent_memory_request: "256Mi"
+  # opflex_agent_memory_limit: "5Gi"
+  acc_provision_operator_memory_request: "256Mi"
+  acc_provision_operator_memory_limit: "5Gi"
+  # aci_containers_operator_memory_request: "256Mi"
```
snaiksat, line 88: Are L85 and L88 intentionally commented out? If so, how would the setting of those values get tested?

pkharat: > Are L85 and L88 intentionally commented out? 

Yes, intentionally commented to test the setting of aci_container_memory_request /  aci_container_memory_limit.

> If so, how would the setting of those values get tested?

Kind of relying on other container request, limit flags for this testing

Do we want to remove these commented variables ?

snaiksat: As a part of this PR, I would prefer that we uncomment those so that we at least end up testing all the overrides consistently. In a follow up PR we can perhaps add variations of the test which exercise parts of the configuration.

pkharat: okay

snaiksat (approved):

PR #984, opened 2023-07-14
Support for k8s 1.26 flavor

akhilamohanan:
Added support for k8s 1.26 flavor


akhilamohanan:
Handled in https://github.com/noironetworks/acc-provision/pull/988

PR #987, opened 2023-07-17
Cluster Report : Collect all available log files of a pod (kmr2)

pkharat:


snaiksat:
Currently not backporting, can reopen if required.

PR #1013, opened 2023-08-15
Fix for getting correct user for travis build

abhis2112:


.travis.yml
```diff
@@ -15,6 +15,7 @@ jobs:
         - python provision/setup.py install
       script:
         - cd provision; coverage run --source=. -m pytest acc_provision; coveralls
+        - echo "Skip running UTs"
```
snaiksat, line 18: this is not correct, we are running UTs in L17, why are we adding this echo?

snaiksat (approved):

PR #1017, opened 2023-08-25
Added changes to get NAD VLAN mapping info from csv file and prepare …

YogeshRajmane:
…NAD VLAN map CR


snaiksat:
Thanks @YogeshRajmane . Can we add a UT which reads a CSV, populates the CR, and then validates it with the expected output?

YogeshRajmane:
> Thanks @YogeshRajmane . Can we add a UT which reads a CSV, populates the CR, and then validates it with the expected output?

Sanjib prepared UT and I added that in same PR.

shastrinator:
Can you please use the latest image and try it on the cluster once..

shastrinator:
Changes look ok

shastrinator:
I have updated the staticfabricattachment CR, so please update in the changeset as well. [https://raw.githubusercontent.com/noironetworks/aci-containers/master/pkg/hostagent/testdata/aci.fabricattachment_staticfabricnetworkattachments.yaml](url)

snaiksat (approved):

PR #1023, opened 2023-09-08
Add support to get vlans for secondary_vlans from ip_sheet_file

YogeshRajmane:
Changes done:
1. Reading VLANs from ip_sheet_file and adding it in secondary_vlans.
2. Modified code for if ip_sheet_file is there then secondary_vlans is optional. if ip_sheet_file is not there then secondary_vlans is mandatory.
3. Fixed unit tests.

Tests done:
1. Verified unit tests.
2. Scenarios verified:
    i. If ip_sheet_file option present and secondary_vlans option absent in input file:
       - secondary_vlans will contain only vlans from ip_sheet_file.
   ii. If ip_sheet_file and secondary_vlans options are present in input file:
       - secondary_vlans will contain union of vlans from ip_sheet_file and already existing secondary_vlans.
  iii. If ip_sheet_file and secondary_vlans options absent in input file:
      - Throwing error and asking user to retry.
  iv. If ip_sheet_file option present and secondary_vlans option absent in input file:
      - secondary_vlans will contain vlans from ip_sheet_file.
   v. If ip_sheet_file option provided but file doesn't exists OR file exist but empty OR file exist but VLANs not there OR file exist but VLANs are not integer value.
      - Throwing error and asking user to retry.


shastrinator (approved):
A range check for the vlanid[0,4095), would be good, but in general, changes look ok.

PR #1027, opened 2023-09-13
Added code to generate FabricVlanPool CR and removed aci-additional-v…

YogeshRajmane:
…lans config from controller and host-agent configmap.

Added code for following:
1. Generating FabricVlanPool CRD and CR.
2. FabricVlanPool will contain VLANs from secondary_vlans and from ip_sheet_file if provided.
3. Fixed Unit Tests.

Tests Done:
1. Verified Unit tests.
2. Verified manifests generation using acc-provision command without APIC.


snaiksat:
@YogeshRajmane thanks. Can we please also move "secondary_vlans" from the net_config section to the chained_cni_config section? This helps to keep all chained CNI related config in one place (since we use secondary_vlans only in that context).

YogeshRajmane:
> @YogeshRajmane thanks. Can we please also move "secondary_vlans" from the net_config section to the chained_cni_config section? This helps to keep all chained CNI related config in one place (since we use secondary_vlans only in that context).

There is one more option called secondary_aep which is also specific to chained mode and it's there in aci_config. I need to move that also.

shastrinator (approved):
LGTM

snaiksat (approved):

PR #1033, opened 2023-10-04
Assisted installer

camrossi:
This adds experimental support for assisted installer on OCP 4.14 (tested with RC1 to 4)
what it does is

1. disable the `installer_provisioned_lb_ip` parameter 
2. add a new RoleBinding and Role
3. a new flavor

**Ingress considerations:**
The way ingress works now with my changes is as following (this is just an idea and can perhaps be changed)
1. Assisted Installer deploys a dedicated set of haproxy, keepalived and coredns pods. These are used to LB the OCP API as well as the default ingress controller.
2. If you want to use ACI PBR for loadbalacning you can create an additional `IngressController` resource as shown below. This will result in having 2xIngress PODs but keeps the control plane cluster connectivity completely disjoint from ACI PBR and I like this approach. 

```
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
  namespace: openshift-ingress-operator
  name: aci-pbr-lb
spec:
  domain: aci-pbr.ocp1nutanix.cam.ciscolabs.com
  endpointPublishingStrategy:
    type: LoadBalancerService
    loadBalancer:
      scope: External
```


snaiksat:
Thanks @camrossi , closing this since this: https://github.com/noironetworks/acc-provision/pull/1058 is merged now.

PR #1047, opened 2023-10-31
Validate openshift system_id

pkharat:


provision/acc_provision/acc_provision.py
```diff
@@ -1194,6 +1194,13 @@ def is_valid_file(path):
     return True
 
 
+def validate_system_id_if_openshift(system_id, flavor):
+    if "openshift" in flavor:
```
snaiksat, line 1198: can we make this case independent (just in case at some point in the future we start writing openshift with upper case characters)

snaiksat (commented):
Thanks @pkharat . Can we add a UT?

provision/acc_provision/acc_provision.py
pkharat: yes, sure

pkharat: Addressed

snaiksat (approved):

PR #1048, opened 2023-11-07
Add ruamel pkg version

pkharat:
ruamel.yaml has released latest version 0.18.5, we have used 0.17.21 for yaml operations(load, edit, write etc). As per release notes, there are changes in the way functions are used/behaving in the latest version. For now fixing ruamel version to 0.17.21, meanwhile will fix unit test framework to adjust to new ruamel version

### DIFF
--- a/provision/setup.py
+++ b/provision/setup.py
@@ -46,6 +46,6 @@ setup(
           'pyopenssl',
           'MarkupSafe',
           'boto3',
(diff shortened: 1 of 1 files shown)


PariyaAshok (approved):

PariyaAshok:
fix-testdata.sh script works only with ruamel version 0.17.21.

PR #1060, opened 2023-12-06
Add a new parameter enable_opflex_agent_reconnect

shareefoc:


provision/acc_provision/templates/provision-config.yaml
```diff
@@ -142,6 +142,7 @@ registry:
   # aci_containers_operator_memory_limit: "5Gi"         # override if needed, default is aci_containers_memory_limit
   # toleration_seconds: 100 # set tolerationSeconds value for controller pod, default value is 60 and 600 for non rke and rke flavors respectively.
   # opflex_openssl_compat: True                         # default is False, use when turning of FIPS on the host
+  # enable_opflex_agent_reconnect: false                # default is False, set it as True to enable opflex agent reconnect after vm migration
```
snaiksat, line 145: nit - might have been good to set this to true in the example (since default is false); perhaps we can do it as a part of later change

PR #1067, opened 2024-01-03
Added RKE-1.4.x and 1.5.x flavors.

jeffinkottaram:
New variables:

 kube_config:
- opflex_agent_statistics
- add_external_contract_to_default_epg
- enable_opflex_agent_reconnect
- opflex_openssl_compat
- node_snat_redirect_exclude
- toleration_seconds

### DIFF
--- a/provision/testdata/list_flavors.stdout.txt
+++ b/provision/testdata/list_flavors.stdout.txt
@@ -30,6 +30,7 @@ INFO: kubernetes-1.25:	Kubernetes 1.25
 INFO: RKE2-kubernetes-1.26:	Rancher Kubernetes Engine Government kubernetes version 1.26
 INFO: RKE2-kubernetes-1.25:	Rancher Kubernetes Engine Government kubernetes version 1.25
 INFO: RKE2-kubernetes-1.24:	Rancher Kubernetes Engine Government kubernetes version 1.24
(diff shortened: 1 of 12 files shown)


provision/testdata/list_flavors.stdout.txt
```diff
@@ -30,6 +30,8 @@ INFO: kubernetes-1.25:	Kubernetes 1.25
 INFO: RKE2-kubernetes-1.26:	Rancher Kubernetes Engine Government kubernetes version 1.26
 INFO: RKE2-kubernetes-1.25:	Rancher Kubernetes Engine Government kubernetes version 1.25
 INFO: RKE2-kubernetes-1.24:	Rancher Kubernetes Engine Government kubernetes version 1.24
+INFO: RKE-1.5.2-experimental:	Rancher Kubernetes Engine min version 1.5.2
+INFO: RKE-1.4.13-experimental:	Rancher Kubernetes Engine min version 1.4.13
```
snaiksat, line 34: "experimental" should not be in the name of the flavor, it should be in the description, see L42 below. That way we can just update the description later to remove the "experimental" qualification. The way it stands now, the user will have to change the flavor name from experimental to without it (essentially its a new flavor).

snaiksat (approved):

PR #1068, opened 2024-01-09
remove ruamel package dependency.

PariyaAshok:


PariyaAshok:
Note:
To run fix-testdata script, We need to install below ruamel version.
pip install ruamel.yaml==0.17.21

snaiksat:
We are actually not removing the dependency, we are just removing its pinning to a specific version.

snaiksat (approved):

PR #1072, opened 2024-01-11
Revert "Bump release to 6.0.3.3"

anmol372:
Reverts noironetworks/acc-provision#1070, due to some unintended changes in testfiles


snaiksat (approved):
Approving since this seems to be creating issues with UTs in some environments (although Travis seems to have passed for this earlier, and we need to investigate how that happened).

PR #1085, opened 2024-01-31
Acc provision unit tests fix to use upstream ruamel version

pkharat:


provision/acc_provision/test_main.py
```diff
@@ -2060,17 +2060,18 @@ def compare_tar_plaintext(expected, output, debug, generated, cleanupFunc):
         tar.close()
 
         # 4 Compare generated and expected *-ConfigMap-aci-operator-config.yaml and other files in the tar folder
```
abhis2112, line 2062: provide this comment just above the  line ```assert prepare_gen_yamls_list == prepare_exp_yamls_list, cleanupFunc()```

abhis2112 (commented):
Request Changes

provision/acc_provision/test_main.py
pkharat: @abhis2112 the thought behind keeping the comment at this place was as the comment is for entire section(including processing) where we are verifying the tar folder files as well.

abhis2112 (approved):

PR #1095, opened 2024-02-26
Remove install_istio flag

smshareef:
Commented out code for disabling the install_istio flag. Functionality is disabled to remove dependency from istio.io/istio package. Vulnerabilties were detected by quay.io security scan of aci-containers-controller and aci-containers-operator images for istio.io/istio package


provision/acc_provision/acc_provision.py
```diff
@@ -149,7 +149,7 @@ def prepare_nadvlanmap(file_path):
                 all_resources[resource_header].append(resource_item)
     except Exception as ex:
         print("Error while preparing yaml contents from given CSV %s"
-            " file. Error: %s" % (file_path, ex))
+              " file. Error: %s" % (file_path, ex))
```
akhilamohanan, line 152: is this change required?

snaiksat:
Thanks @smshareef . The commit message is a bit misleading. The install_istio config is already disabled by default. What we are taking away is the ability to enable it.

Disabling the configuration to enable install_istio to remove dependency from istio.io/istio package since this feature will not be supported going forward. For the past few cycles we have moved towards supporting Red Hat Service Mesh operator.

provision/acc_provision/templates/provision-config.yaml
```diff
@@ -157,7 +157,6 @@ registry:
     #enable_updates: True # Default is False
 
 #istio_config:
-  # install_istio: False         # default is True
   # install_profile: "default"   # override if needed, default is "demo"
```
snaiksat, line 160: this whole section needs to be removed

snaiksat (changes requested):

provision/acc_provision/templates/provision-config.yaml
smshareef: Done, thanks for the review @snaiksat.

akhilamohanan (approved):

PR #1112, opened 2024-04-03
Skip openshift flavor system_id validation during cluster upgrade (--upgrade)

pkharat:


snaiksat (approved):

snaiksat:
@pkharat thanks; as a follow up can we please add a couple of UTs?

PR #1113, opened 2024-04-10
Removed default value of lldpDiscovery option from NFC CRD

YogeshRajmane:


shastrinator (approved):
LGTM. This is based on a discussion with Cam.

PR #1114, opened 2024-04-16
Changes for local hpp distribution

smshareef:


provision/acc_provision/templates/provision-config.yaml
```diff
@@ -159,6 +159,7 @@ registry:
   # add_external_contract_to_default_epg: True          # override if needed, default is False
   # apic_connection_retry_limit: 5                      # number of times the controller tries to communicate with APIC before switching to next APIC if unsuccessful, default is 5
   # taint_not_ready_node: True                          # default is False, set to True if you want to make the node in not ready state unschedulable till the host agent initalization is complete.
+  # enable_local_netpol: True                           # default is False, set to True to enable local hpp distribution
```
snaiksat, line 162: can we change the name to: enable_hpp_direct
and description to say: default is False, set to True to enable HPP distribution via Kubernetes control plane for faster convergence and reduce load on fabric

snaiksat (approved):

PR #1115, opened 2024-04-22
Provision APIC Service Graph Resources

smshareef:


provision/acc_provision/apic_provision.py
```diff
@@ -8560,6 +8563,684 @@ def bgp_peer_prefix(self):
         self.annotateApicObjects(data)
         return path, data
 
+    def service_graph_template(self):
+        print("INFO: Creating Service Graph Template Object", file=sys.stderr)
+        system_id = self.config["aci_config"]["system_id"]
+        path = "/api/node/mo/uni/tn-common/AbsGraph-%s_svc_global.json" % (system_id)
+        dn = "uni/tn-common/AbsGraph-%s_svc_global" % (system_id)
```
akhilamohanan, line 8570: we should not hardcode common. It is aci_config->vrf->tenant

snaiksat:
thanks @smshareef , please add a commit message so that its clear what this patch is intended for...

PR #1118, opened 2024-05-09
Add document for opflex agent ERSPAN feature

pkharat:


snaiksat:
thanks @pkharat , couple of comments inline; can we also pl cross-reference this from configurations/features list?

pkharat:
> thanks @pkharat , couple of comments inline; can we also pl cross-reference this from configurations/features list?

In the original doc, which parameter to use in acc provision input file to enable erspan in k8s is not mentioned, this one https://github.com/noironetworks/acc-provision/blob/opflex-erspan-doc/provision/acc_provision/templates/provision-config.yaml#L179 Should we add information about this ?

snaiksat:
thanks @pkharat , yes please include this configuration in the doc

PR #1146, opened 2024-06-28
Reduce the size of acc-provision input pushed to APIC

tomflynn:
We push the input for acc-provision to vmmInjectedClusterInfo.accProvisionInput. The max length supported by the APIC for the field is 8192. Some users are hitting this limit with the increased number of comments and whitespace in the input file.
Strip out comments and whitespace by loading/dumping the yaml file before POST


snaiksat:
@pkharat , I think we might be running into some UT problems here similar to the ones you had hit earlier when you were parsing yaml for the UT base-decoding patch. Can you please take a look to see what could be wrong here?

pkharat:
> @pkharat , I think we might be running into some UT problems here similar to the ones you had hit earlier when you were parsing yaml for the UT base-decoding patch. Can you please take a look to see what could be wrong here?

@snaiksat one UT  "test_base_case_apic_5_2_3" was failing in the PR  reason is it was not present in fix-testdata.sh, so the expected apic.txt was not modified in the testdata. I added the test in fix-testdata.sh https://github.com/noironetworks/support/pull/3043/files. Re-running this updated fix-testdata.sh on this branch should solve this.

snaiksat (approved):

PR #1150, opened 2024-07-09
Remove EPG+contract CRDs

tomflynn:
These were only ever supported with overlay cases

Remove netpol+dns policies for overlay case as well as they aren't used/supported


tomflynn:
https://github.com/noironetworks/aci-containers/pull/1335 needs to merge first

snaiksat (approved):

PR #1156, opened 2024-08-02
Preprovisioning svi changes

YogeshRajmane:


shastrinator (approved):
Thanks for catching up on some missing status fields.

PR #1210, opened 2024-11-24
add rke flavor 1.6.5 and 1.7.1

akhilesh-oc:


provision/testdata/list_flavors.stdout.txt
```diff
@@ -44,11 +44,13 @@ INFO: kubernetes-1.28:	Kubernetes 1.28
...
+INFO: RKE-1.7.1:	Rancher Kubernetes Engine min version 1.7.1
+INFO: RKE-1.6.5:	Rancher Kubernetes Engine min version 1.6.5
+INFO: RKE-1.6.3:	Rancher Kubernetes Engine min version 1.6.3
+INFO: RKE-1.6.2:	Rancher Kubernetes Engine min version 1.6.2
 INFO: RKE-1.6.0:	Rancher Kubernetes Engine min version 1.6.0
-INFO: RKE-1.5.14:	Rancher Kubernetes Engine min version 1.5.14 [Experimental]
-INFO: RKE-1.5.13:	Rancher Kubernetes Engine min version 1.5.13 [Experimental]
+INFO: RKE-1.5.14:	Rancher Kubernetes Engine min version 1.5.14
+INFO: RKE-1.5.13:	Rancher Kubernetes Engine min version 1.5.13
 INFO: RKE-1.5.11:	Rancher Kubernetes Engine min version 1.5.11
```
snaiksat, line 54: we are removing "experimental" from a bunch of flavors, and adding new flavors without experimental, are the RKE releases for all these flavors already available?

also, shouldn't we be hiding some of the older flavors, at least say 1.3.x, 1.4.x?

akhilesh-oc: Yes I will hide older flavors.
Regarding new Experimental tags, As rancher has already confirmed the upcoming release versions, we directly use w/o experimental tags.

jeffinkottaram: The RKE releases for all the flavors for which experimental tag was removed are available and tested.

snaiksat (approved):

PR #1245, opened 2025-03-04
Delete AP and EPGs which are created by acc-provision for a specific system-id

Santosh-oc:
- Improve system-id match using substring pattern
- Check if annotation is present before deleting resource
- Delete application profile only when all its children EPGs are annotated else do not delete that application profile

### DIFF
--- a/provision/acc_provision/apic_provision.py
+++ b/provision/acc_provision/apic_provision.py
@@ -152,6 +152,13 @@ class Apic(object):
 
         err("Max retries reached for POST request to %s. Giving up." % path)
         return resp
+    
+    def is_system_id_matching(self, system_id, resource_name):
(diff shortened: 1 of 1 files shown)


provision/acc_provision/apic_provision.py
```diff
@@ -152,6 +152,13 @@ def post_with_exponential_backoff(self, path, data, max_retries=5):
 
         err("Max retries reached for POST request to %s. Giving up." % path)
         return resp
+    
+    def is_system_id_matching(self, system_id, name):        
+        exact_match_pattern = rf".*-\b{system_id}\b-.*"
```
snaiksat, line 157: nit: while its just a variable name, the regex is not for an exact match, right? its more like "contains_match_pattern"...

pkharat (approved):

PR #1252, opened 2025-03-17
Unit Test for deletion of APs and EPGs

Santosh-oc:
- Ensure deletion of APs and EPGs created by acc-provision for a specific system-id.
- Validate that APs and EPGs are deleted based on matching system-id and annotation.

### DIFF
--- a/provision/testdata/test_unprovision_custom_ap_with_annotation.inp.yaml
+++ b/provision/testdata/test_unprovision_custom_ap_with_annotation.inp.yaml
@@ -1,0 +1,135 @@
+aci_config:
+  system_id: kube
+  #use_legacy_kube_naming_convention: False
+  tenant:
+    name: test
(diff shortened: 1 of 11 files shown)


provision/testdata/test_unprovision_custom_ap_with_annotation.inp.yaml
```diff
@@ -0,0 +1,135 @@
+aci_config:
+  system_id: kube
+  #use_legacy_kube_naming_convention: False
+  tenant:
+    name: test
+  apic_hosts:                   # List of cAPIC hosts to connect for APIC API
+  - localhost:50001
+  apic_version: "5.1"
```
snaiksat, line 8: not sure how relevant it is, but lets change this to 6.1, and correspondingly lets make sure that the json responses from the APIC are based on the latest 6.1

akhilamohanan (approved):

